PR #21, opened 2026-07-21
fix: P0 correctness batch 2 (embed sentinel, learn atomicity, embed-outside-tx, download hardening, SSH host keys, Gemini cache)

knomit:
Second batch of P0 correctness fixes from the 2026-07-20 code review. Base is `dev`; this is the bottom of a three-PR stack (P0-batch-2 → P0.4/P1.1 → P1.2–P1.7).

Each fix diverged from the report where the report's suggestion was wrong; the divergences are the interesting part.

- **P0.2 — web embed sentinel.** Neither suggested fix used: `emptyOutDir:false` would accumulate stale hashed assets into the embedded FS, and a Makefile re-touch leaves a bare `npm run build` still breaking it. A vite plugin snapshots `.gitkeep` at config-eval time and rewrites it in `closeBundle`, so emptying still happens but the sentinel survives any entry point.
- **P0.3 — learn subsume atomicity.** `BatchWriteFacts` gained a `deletes` parameter so a write and its retraction share one commit; the swallowed `DeleteFact` error is gone. The running root is seeded from the parent tree so a delete-only batch is valid.
- **P0.5 — ONNX embed out of the write tx.** "Embed first" alone would cost an inference on every re-index, so a read-only COW probe runs before the tx and decides whether an embedding is needed; the authoritative in-tx COW check is unchanged, and a probe invalidated by a concurrent writer falls back to embedding in-tx rather than committing a vectorless fact.
- **P0.6 — model download hardening.** Not a blunt `Client.Timeout` (would kill legitimate slow-link downloads of a hundreds-of-MB model) — bounded dial/TLS/header phases plus a 90s stall watchdog. Integrity via Content-Length (catches the truncated-200 failure); SHA-256 pins wired but left unset because the registry points at mutable HuggingFace `main`. `tools/fetchlibs` got temp-file-then-rename.
- **P0.7 — SSH host-key verification.** `known_hosts`-backed TOFU callback: unknown host pins, known host with a different key rejects. An unusable `known_hosts` is a hard error rather than a silent fall-back to insecure.
- **P0.8 — Gemini cache TTL.** Cache entries store expiry; a stale entry is treated as a miss and evicted; a cache-missing error retries inline but only when nothing was emitted, so a retry can't duplicate output.

Regression-anchored throughout. `go test ./...` green.

### DIFF
--- a/cmd/warm.go
+++ b/cmd/warm.go
@@ -37,7 +37,7 @@ func warmModelsCmd() *cobra.Command {
 				return fmt.Errorf("unknown embedding model %q: %w", id, err)
 			}
 			cacheDir := filepath.Join(cfg.Home, "models")
-			modelPath, tokPath, err := embeddings.EnsureModel(model, cacheDir)
+			modelPath, tokPath, err := embeddings.EnsureModel(cmd.Context(), model, cacheDir)
 			if err != nil {
 				return fmt.Errorf("download model %q: %w", id, err)
 			}
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -70,7 +70,7 @@ func New(ctx context.Context, cfg config.Config, opts Options) (*App, error) {
 	if err != nil {
 		return nil, fmt.Errorf("embedder model config invalid (embeddings.model=%q): %w", cfg.Embeddings.Model, err)
 	}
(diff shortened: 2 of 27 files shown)


knomit:
### Code review

PR #22, opened 2026-07-21
refactor: extract the shared synthesis engine + atomic work-item claim (P0.4, P1.1)

knomit:
P0.4 (review pipeline idempotency) and P1.1 (shared synthesis engine) from the 2026-07-20 code review. **Stacked on #21** — review that first; this PR's diff is against `fix/p0-batch-2`.

The two items landed together because the engine extraction is what makes the idempotency fix clean.

**P0.4 — atomic work-item claim, and a third ordering the report didn't consider.** The report framed the choice as apply-then-mark vs mark-then-apply. Neither is right: the first duplicates synthesized facts on retry, the second silently loses work. The order is now **peek → decode+validate → CAS-claim → apply**. Pulling validation before the claim is the load-bearing move — malformed LLM JSON (the common failure) leaves the item untouched and fully retryable, so the claim only ever consumes a well-formed response. `SetPipelineWorkItemResponse` became `AnswerPipelineWorkItem`, whose `(claimed bool, err error)` makes a lost race a benign no-op.

Accepted and documented rather than fixed: a hard failure *during* apply loses that item's decisions — deliberate, because the corpus is left un-maintained, never invalid, and a lease+retry would reintroduce the duplication it exists to prevent (apply isn't idempotent — `decision.go` mints a fresh UUID per apply).

**P1.1 — one engine, two strategies.** The KB decision `one-engine-two-drivers` said the Reviewer was the single engine; in reality hypothesize reimplemented the whole session lifecycle inside the MCP transport layer, and the two copies had already made *opposite* idempotency choices. `synthesize.Pipeline` now owns session lifecycle, phase machine, seed scan + watermark, completion, and the P0.4 claim protocol; a `Strategy` interface is the per-tool surface. `review.go` went 1258 → 190 lines, `mcp/hypothesize.go` 624 → 141. Both strategies are zero-size structs, so "no session state" is enforced by the type system. Acceptance criterion: the `internal/synthesize` suite passed unedited.

Three latent bugs the extraction fixed for free: hypothesize was violating `session-branch-binding`; its defensive re-read of the scoped flag is gone; and the seed projection now carries `Refs`, which the old path dropped. `chunkFacts` wired (64 KiB, distill only), dead helpers deleted, `Summary` populated, hypothesize got TaskSupport and lost its 30s timeout.

`go test ./...` green.


knomit:
### Code review

PR #28, opened 2026-07-22
refactor: remove the GraphQLite dependency

knomit:
Removes the GraphQLite engine. The property graph becomes plain SQLite tables that knomit owns and queries directly.

Three commits, each building and testing standalone:

1. **`feat(store)`** — direct-SQL graph write primitives (additive; engine untouched)
2. **`refactor(store)`** — port all 27 `cypher()` sites to direct EAV SQL (engine loaded but unused)
3. **`refactor`** — remove the engine and its build-pipeline footprint

## Why

We were already routing around the engine: it can't parameterise writes (so every write was string-interpolated Cypher), it silently mis-executes multi-MERGE, and `cypher_retry.go` existed purely to absorb its translation races. Meanwhile the DERIVED_FROM path already bypassed it with direct EAV SQL.

An audit of all 27 call sites found **zero multi-hop Cypher traversals** — every read is 1-hop or a label scan, and the one real transitive walk (blast radius) was already a recursive CTE. gonum does the actual graph algorithm (Louvain). Nothing was riding on the query engine.

## The load-bearing discovery

The graph tables were **never declared anywhere**. The extension created them lazily, triggered by migration `000003` being literally `SELECT cypher('RETURN 1')`. Removing the engine naively leaves a fresh database with no graph at all.

Migration `000014` now declares them explicitly, matching the layout the extension used, with `IF NOT EXISTS` so migrated databases no-op. `000003` becomes a no-op but is **retained** — golang-migrate tracks applied versions by number, so deleting it would strand databases on a version that no longer exists.

## Forced rebuild (one-time)

`GraphSchemaVersion` 4 → 5. Existing databases hold `deleted` as a JSON boolean in `node_props_bool`, which the direct-SQL readers cannot see — without the bump every node would read as live. Rebuild regenerates the graph from git and **preserves embeddings** (`TestRebuild_PreservesRowidsAndDoesNotReEmbed`), so this costs a graph rewrite, not a re-embed.

Properties are now TEXT throughout. No graph read does a numeric or range comparison; `confidence`/`sources` are written but never read back, and `deleted` reduces to `'true'`/`'false'`. The typed property tables (`_int`/`_real`/`_bool`/`_json`) are dropped.

## Also removed

The extension load, `graphqliteLibPath`/`dirEvalSymlinks`, `cypher_retry.go` and its metric, `escapeCypherKey`/`escapeCypherVal`, `isDeletedVal`, `jsonParams`, and the pre-warmed connection pool — that pool existed only because a lazy connection open ran the ConnectHook, which loaded the extension and wrote to its EAV schema, risking self-deadlock under `_txlock=immediate`. The hook now only registers SQL functions and sets pragmas.

**`_txlock=immediate` is deliberately kept.** Relaxing it was measured to buy essentially nothing: reads already run in autocommit and never take the write lock, WAL already gives reader/writer concurrency, and SQLite is single-writer regardless.

Build side: graphqlite drops out of `tools/fetchlibs`, the Makefile, Dockerfile, CI and release workflows, the desktop bundle and `install.sh`. The codesign machinery goes too — graphqlite was its only user, shipping unsigned dylibs Gatekeeper refused to `dlopen`.

## Two intentional behaviour changes

- **`gcOrphanedGraphNodes` had a latent bug.** It projected `n.path`, but Entity nodes are keyed by `name` and have no `path`, so entity orphans were **never actually collected**. Matching by node id fixes it. It's also one set-based DELETE instead of a round-trip per orphan (a P2 item, free).
- **Injection-shaped paths are now parameterised** rather than escaped into a Cypher literal. The injection tests still pass, now proving the parameterised form.

## Review notes

- **SIMILAR_TO is stored directed but read undirected.** The ported SQL unions both orientations — querying one direction would silently shrink every cluster. Worth a careful look in `graphSimilarToNeighbours`.
- **Writes and reads of `deleted` had to move together**; porting one without the other mis-filters silently. This surfaced as a real test failure during development.
- **Post-commit DERIVED_FROM is left as-is.** Its original constraint (direct-SQL reads couldn't see Cypher-MERGE'd nodes in the same transaction) is genuinely gone now, so it *could* move into the transaction — deliberately not done here, to keep this behaviour-neutral. Noted in the code as a follow-up.
- Stale Cypher-era comments were corrected rather than left to mislead, including the subgraph chunk size — its real limit is now SQLite's host-parameter cap, not GraphQLite's expression-tree depth.

## Testing

**The full repo build, vet, and every test package pass with the extension no longer loaded** — the real proof, since tests create fresh databases and exercise the graph end to end.

New tests pin the schema as ours: migration `000014` creates the tables on a fresh database, the typed tables stay dropped, and deleting a node still cascades labels, properties and edges in both directions (the invariant behind reducing `DETACH DELETE` to `DELETE FROM nodes`).

## Not in scope

Typed edge tables (originally "Stage 3") stay deferred. Measurement on a live corpus shows the EAV schema resolves every join through covering indexes — 556 fact-node lookups in 2 ms, a 608-node recursive walk in 5 ms — so there is no performance case today, and no roadmap feature needs a richer graph shape.


knomit:
### Code review

Found 6 issues. All are in comments/tests, not shipped behaviour — the port itself held up under a history pass, a comment pass, and a bug pass. Four of the six are the PR's own last two commits (`414ac177` tx-wrapping, `d12dc919` migration 000015) invalidating claims that earlier commits in the same PR wrote.

1. `graphMergeEdge`'s justification is falsified by this PR's own later commits: no caller uses the bare pool any more (`414ac177` wrapped `graphBuildSimilarityEdges` in `beginTxIfNeeded`), and `edges` now *does* have a uniqueness constraint for every type except DERIVED_FROM (migration `000015`, whose own comment says the index "makes the bug impossible to reintroduce").

https://github.com/knomit/knomit/blob/d12dc919d7f15c0099b15de7f80e6232d06c710c/internal/store/graph_sql.go#L197-L204

2. Same two false claims in the test's doc comment, contradicted by `TestEdges_MergeIdentityIndexIsEnforced` 290 lines below in the same file, which asserts the schema *does* reject a duplicate SIMILAR_TO insert. The "HONEST LIMIT" note is also moot now — a split two-statement form would fail on constraint violation, not duplicate silently.

https://github.com/knomit/knomit/blob/d12dc919d7f15c0099b15de7f80e6232d06c710c/internal/store/graph_sql_test.go#L140-L150

3. `graphInsertEdgeReturningID`'s doc states "(source_id, target_id, type) has no uniqueness constraint … this function always produces a new row". After `000015` that holds only for DERIVED_FROM. No live bug — the sole caller passes `EdgeDerivedFrom` — but the doc invites a future caller to use it for a relationship edge, which would now hit a UNIQUE failure the comment says cannot happen.

https://github.com/knomit/knomit/blob/d12dc919d7f15c0099b15de7f80e6232d06c710c/internal/store/edge_props.go#L9-L14

4. Third copy of the "direct-SQL reads cannot see Fact nodes MERGE'd via Cypher inside the same `*sql.Tx`" rationale, missed by the sweep. The PR correctly rewrote the identical claim in [`derived_from.go`](https://github.com/knomit/knomit/blob/d12dc919d7f15c0099b15de7f80e6232d06c710c/internal/store/derived_from.go#L216-L223) and [`search_crud.go`](https://github.com/knomit/knomit/blob/d12dc919d7f15c0099b15de7f80e6232d06c710c/internal/store/search_crud.go#L358-L365) ("that constraint is gone … left as-is to keep the removal behaviour-neutral"), so a reader here is told the post-commit phase is technically forced when it is only a deliberate hold-back. `verify.go:755`/`759` also still say "graphqlite Fact nodes" / "the graphqlite model".

https://github.com/knomit/knomit/blob/d12dc919d7f15c0099b15de7f80e6232d06c710c/internal/store/search_index.go#L1149-L1153

5. `TestRebuildGraph_SimilarityRespectsTopKCap` cannot fail. It writes 5 facts, so max out-degree is 4; `knnK` is 10 ([`search_graph.go:313`](https://github.com/knomit/knomit/blob/d12dc919d7f15c0099b15de7f80e6232d06c710c/internal/store/search_graph.go#L312-L314)); and `000015` makes duplicates schema-impossible. `require.LessOrEqual(maxOut, knnK)` is satisfied by every conceivable behaviour including `maxOut == 0`, since there is no positivity floor. Measured at head: `maxOut=4, knnK=10`. The regression it documents (out-degree 28→34 by accumulation) needs more than `knnK+1` facts plus a pre-seeded stale surplus.

https://github.com/knomit/knomit/blob/d12dc919d7f15c0099b15de7f80e6232d06c710c/internal/store/rebuild_similarity_prune_test.go#L114-L116

6. `414ac177`'s atomicity fix shipped with no test (`1 file changed, 32 insertions(+), 2 deletions(-)`, zero test files), so splitting the transaction back apart would fail nothing. Relatedly `inSimTx` is dead: repo-wide it appears only at its declaration and its fire site (`search_index.go:1372`) — only `beforeSimTx` is used, by `rebuild_similarity_race_test.go:77`.

https://github.com/knomit/knomit/blob/d12dc919d7f15c0099b15de7f80e6232d06c710c/internal/store/search_index.go#L66-L70

---

### Upgrade-path risk (below the reporting bar, but worth a decision)

Migration `000015` creates the unique index with no prior dedup. The mechanism was verified end to end: `migrate.All` runs inside `store.Open` ([`service.go:127`](https://github.com/knomit/knomit/blob/d12dc919d7f15c0099b15de7f80e6232d06c710c/internal/store/service.go#L126-L128)), well before the `GraphSchemaVersion` 4→5 heal, which needs an opened store. A failing `CREATE UNIQUE INDEX` leaves `schema_migrations.dirty = 1`; there is no `Force(` or `ErrDatabaseDirty` handling anywhere in the tree, so a non-default repo is dropped behind a single `Warn` and the *default* repo fails server start outright. The forced rebuild that would regenerate a clean graph can never run.

The file's comment says failing loudly "is intended — a one-off audit". That intent is defensible, but what ships is not an audit; it is an unrecoverable database with no documented remedy. What I could not establish is how plausible duplicate edges actually are in a real v4 database — Cypher `MERGE` writes went through the extension rather than the caller's `*sql.Tx`, so `_txlock=immediate` did not cover its internal check-then-insert, and shared-target edges (`CHILD_OF`, `TAGGED`) are the realistic collision candidates; but I found no commit, test, or comment reporting an observed duplicate.

A pre-index `DELETE FROM edges WHERE type <> 'DERIVED_FROM' AND id NOT IN (SELECT MIN(id) FROM edges GROUP BY source_id, target_id, type)` costs nothing: `graphSetEdgeProps` has exactly one caller and it is DERIVED_FROM-only, so non-DERIVED_FROM edges carry no properties, and version 5 regenerates them regardless.

https://github.com/knomit/knomit/blob/d12dc919d7f15c0099b15de7f80e6232d06c710c/internal/store/migrate/migrations/000015_edges_merge_identity.up.sql#L14-L19

### Against the hardening checklist

`.claude/plans/2026-07-23-graph-hardening-test-checklist.md` was applied as a rubric. What it endorses: `TestMigration000014_ConvertsDeletedBeforeDroppingTypedTables` asserts the migrated state *before* any rebuild; `TestRebuild_DoesNotDestroyConcurrentCrossBranchSimilarityEdges` uses a rendezvous hook rather than sleeps, asserts the hook fired, and carries a positivity floor; `Rebuild` bumps `graph_schema_version` only after the similarity tx commits, so a partial rebuild leaves `NeedsRebuild` true. Residue:

- **Positivity floors missing** where the same PR gets it right elsewhere. `TestRebuildGraph_SimilarityIsIdempotent` compares `first` against a second read with no `require.NotEmpty(first)` — both `nil` passes, and a missing embedder skips the phase entirely. `TestRebuildGraph_PrunesStaleSimilarityEdges` is partly defended by its ghost assertions, but its `baseSimilar`/`baseDerived` checks are satisfied by an empty layer. Compare [`rebuild_similarity_race_test.go:68`](https://github.com/knomit/knomit/blob/d12dc919d7f15c0099b15de7f80e6232d06c710c/internal/store/rebuild_similarity_race_test.go#L67-L69).
- **The upgrade path is fixture-only**, which the checklist explicitly says does not count. `openMigratedTestDB` runs `migrate.All` *first*, so 000014/000015 have already created the final layout; `seedV4Graph` then hand-builds `node_props_bool` and rolls `schema_migrations` back to 13. The replay runs against the new code's own tables, so a column-shape difference in what the extension actually wrote is structurally undetectable. No test opens a v4-stamped database through `store.Open`.
- **Uncovered concurrency cases** from the checklist's list: retract-while-read, rebuild-while-read, two concurrent rebuilds on different branches (both running the superseded-versions prune against the shared graph), and merge-while-rebuild. `test/storytests/concurrency_test.go` still has each branch writing its own path.

https://github.com/knomit/knomit/blob/d12dc919d7f15c0099b15de7f80e6232d06c710c/internal/store/migration_graph_eav_test.go#L32-L36

Checked and cleared: SIMILAR_TO's directed-write/undirected-read union and both callers' `requireAnchorLive` flags; the `deleted` TEXT conversion ordering in 000014; `graphMergeNode`'s check-then-insert (every production caller is inside `BEGIN IMMEDIATE`); the scoped similarity prune's cross-branch safety; `warmPool` removal against `78675d98`'s original rationale; every deleted Cypher-era guard against its introducing commit; `rows.Err()` on all new readers; and the `SimilarityAdjacency` IN-list, which raises the failure threshold ~65x versus the OR-chain it replaces.

knomit:
### Review follow-up — all six addressed in a74c9da8

**Comments (1–4).** Rewrote the `graphMergeEdge` / test / `graphInsertEdgeReturningID` cluster to say "single statement plus `ux_edges_merge_identity`" instead of "no uniqueness constraint, bare pool"; `graph_sql_test.go` no longer contains two tests asserting opposite facts about the schema. Fixed the third copy of the GraphQLite in-tx rationale in `rebuildGraph`, the two `graphqlite` mentions in `verify.go`, the `graph_sql.go` header's "DERIVED_FROM is never pruned" overstatement, the garbled `graphInsertEdge` sentence, and the "ONNX + ONNX" typo.

**Top-K test (5).** Now uses a corpus above the bound and seeds a surplus before the rebuild under test, so it can fail. Two findings while fixing it:

- **The bound is `knnK+1`, not `knnK`.** The KNN asks sqlite-vec for `knnK+1` rows so the fact's own row can be filtered out afterwards; when ties push that self-row outside the returned set, all 11 are neighbours. Measured `maxOut = 11` on a 14-fact corpus. The old `<= knnK` assertion was asserting an invariant the implementation does not have — it only looked true because 5 facts cap out-degree at 4. Asserting the real bound rather than changing behaviour, since the similarity layer is not this PR's to alter.
- **The prune test's DERIVED_FROM assertion was comparing 0 against 0** — its facts carried no refs, so no DERIVED_FROM edges existed to preserve. They now cite each other.

**Atomicity (6).** Added `inSimEdgeTx` alongside `beforeSimTx`/`inSimTx` as the observation point on the incremental path, and tests asserting both in-transaction windows are invisible to a second connection. Reverting 414ac177's transaction wrapping makes the observer see 1 edge where it must see 3 — the thinned neighbourhood that commit describes.

**Migration 000015.** Now dedupes before creating the index. Detail and the interaction with #33 are in the commit message and [a note on that issue](https://github.com/knomit/knomit/issues/33#issuecomment-5060029928).

Every new test negative-controlled: remove the fix, confirm it fails. Full suite green (`go build ./...`, `go vet ./...`, all packages).

**Still open, deliberately.** The upgrade path is still fixture-only — `seedDuplicateEdges` follows the same shape as `seedV4Graph`, replaying against tables the new code already created. It proves the dedupe SQL is correct, not that a real v4 `edges` table matches the assumed layout. That needs a `VACUUM INTO` copy of a real pre-migration database, which is harness work rather than CI work. The uncovered concurrency cases from the checklist (retract-while-read, rebuild-while-read, two concurrent rebuilds, merge-while-rebuild) are also unaddressed.

PR #34, opened 2026-07-23
fix(store): recover from an interrupted migration instead of dropping the repo

### LINKED ISSUE #33: An interrupted migration silently drops the repo from the API, with no recovery path
Found while hardening PR #28. Pre-existing, not caused by that PR.

## What happens

golang-migrate marks `schema_migrations.dirty = 1` before running a migration body and clears it after, in **separate transactions** from the body itself. A crash, SIGKILL, or power loss anywhere in that span leaves `dirty = 1` permanently.

knomit has no dirty detection and never calls `Force()` — `grep -rn "Force(\|dirty" internal/store cmd --include="*.go"` finds nothing relevant (the `anyDirty` in `cmd/verify.go:55` is about verify-report cleanliness, unrelated). `knomit reset` only wipes all data and does not touch `schema_migrations`.

The result is not a crash. The server starts normally and returns HTTP 200 — but **the affected repo is silently dropped**, logged only at WARN:

```
WRN skipping repo error="open store: store.Open: migrate.All: Dirty database version 15. Fix and force version." repo=knomit-kb
```

`GET /api/v1/repos` then returns `["core"]`. From the user's point of view their knowledge base has simply vanished, with a single warning line in the log and no in-product way to get it back.

## Repro

```bash
sqlite3 $KNOMIT_HOME/repos/<repo>.db "UPDATE schema_migrations SET dirty = 1;"
knomit serve
curl -s localhost:<port>/api/v1/repos    # the repo is absent
```

Verified on a copy of a real repo database (version 15, dirty forced to 1). Recovering it by hand requires knowing to run `UPDATE schema_migrations SET dirty = 0` against the right file — not something a user can be expected to discover from the log line.

## Why the data is fine

Worth stating explicitly, because it bounds the severity: golang-migrate wraps the migration **body** in a transaction, so a migration cannot tear. An interrupted migration leaves either the fully-applied or the fully-unapplied schema, never a half-applied one. The database is intact; only the `dirty` bookkeeping flag makes it unopenable.

That is what makes automatic recovery reasonable rather than risky.

## Suggested direction

Since the body is atomic and every knomit migration is written idempotently (`IF NOT EXISTS` / `INSERT OR IGNORE`), a dirty flag at version N can be cleared and the migration re-run safely. Options, roughly in order of preference:

1. Detect `dirty` in `migrate.All` and self-heal: `Force(N-1)` then `Up()`, logging loudly at INFO/WARN that recovery happened.
2. If self-healing is judged too magical, add an explicit `knomit repair --repo <name>` command so there is at least a supported path.
3. At minimum, escalate the log line from WARN to ERROR and make it actionable — name the database file and the exact remedy — and surface the skipped repo in the API rather than omitting it silently, so the failure is visible rather than looking like data loss.

Whichever is chosen, a regression test (`TestMigrate_RecoversFromDirtyVersion`) is worth having, since the failure mode is invisible until someone loses a repo.

## Related

- #32 — the other restore/upgrade-path gap found in the same round.

knomit:
Closes #33.

## The bug

golang-migrate's sqlite3 driver runs **three separate transactions** per migration: `SetVersion(N, dirty=true)`, the body, then `SetVersion(N, dirty=false)`. A crash, SIGKILL, or power loss anywhere across that span leaves `dirty = 1` permanently.

knomit had no dirty handling, so every later open failed, `Manager.Add` skipped the repo, and the user's knowledge base simply vanished from `GET /api/v1/repos` behind a single WARN line — indistinguishable from data loss, with no in-product recovery path.

## The fix

`migrate.All` now self-heals via `upWithRecovery`. The dirty flag does not say which side of the body's commit the crash fell on, so recovery tries both possibilities, **each exactly once**:

1. **Body never landed** — rewind to `N-1` and re-run `N`. This is the case that must work, so it is tried first.
2. **Body landed and `N` is not re-runnable** — only entered when the re-run fails with SQLite's `already exists` / `duplicate column name`, which is itself proof the body committed (the body is transactional, so a partial application would have rolled back). `Force(N)`, then continue.

Anything else is a migration failing **deterministically on the data**. Those re-fail identically forever, and an unguarded loop would clear, re-fail and re-dirty on every boot — turning a one-time manual recovery into a permanent one. So the *underlying* migration error is returned rather than the dirty flag, and `dirty` is left set. This is the guard raised in the issue comment.

## One correction to the issue's premise

The issue states that every knomit migration is written idempotently. **It is not**, and that assumption is load-bearing for step 1:

- `000008_fact_domain_tokens` created its table and index without `IF NOT EXISTS` — **fixed in this PR**.
- `000005`, `000006`, `000011`, `000012` are `ALTER TABLE … ADD COLUMN`, which SQLite cannot express idempotently at all. Step 2 above is what covers them.

## Deliberate non-changes

- **`migrate.Core` does not self-heal.** It targets version 1 explicitly, so rewinding a database dirty at a *later* version would migrate it **down** and destroy schema. Its only callers hand it a fresh `:memory:` database, where the dirty state cannot arise. Documented at the call site.
- **The API surface is unchanged** (option 3 in the issue). A repo that fails to open for some *other* reason is still omitted from `/api/v1/repos` — but the skip site is now ERROR rather than WARN and names the database file, since it remains the only signal.

## Tests

Every test below was confirmed to **fail against the pre-fix behaviour** before being kept.

`internal/store/migrate/recovery_test.go` — white-box, synthetic migration set so each interruption window is reproduced exactly:

- `TestMigrate_RecoversFromDirtyVersion` — window 1, body never committed.
- `TestMigrate_RecoversFromDirtyFirstVersion` — the `Force(0)`-is-invalid edge case; rewinding a dirty *first* migration must go to `NilVersion` or the re-run silently no-ops.
- `TestMigrate_RecoversWhenBodyAlreadyCommitted` — window 2, over an `ALTER TABLE ADD COLUMN`.
- `TestMigrate_DoesNotLoopOnDeterministicFailure` — asserts the underlying error surfaces, `dirty` stays set, and a second boot behaves identically.

`internal/store/migration_dirty_test.go` — the issue's literal repro against the **real embedded migrations**: version unchanged, `dirty` cleared, `ux_edges_merge_identity` intact, second open a no-op.

`go build ./...`, `go vet`, and `go test ./internal/store/... ./internal/repos/...` all pass.

### DIFF
--- a/internal/repos/manager.go
+++ b/internal/repos/manager.go
@@ -535,7 +535,11 @@ func (m *Manager) Start() error {
 			continue
 		}
 		if err := m.Add(name, dbPath); err != nil {
-			log.Warn().Err(err).Str("repo", name).Msg("skipping repo")
+			// ERROR, not WARN: the repo disappears from the API entirely, so
+			// this line is the ONLY signal the user gets. Name the database
+			// file — recovering by hand needs to know which one it is.
+			log.Error().Err(err).Str("repo", name).Str("db", dbPath).
+				Msg("repo failed to open and was skipped; it will not appear in the API")
 		}
 	}
 
--- a/internal/store/migrate/migrate.go
+++ b/internal/store/migrate/migrate.go
@@ -84,49 +84,114 @@ func All(db *sql.DB) error {
 // boot, turning a one-time manual recovery into a permanent one. So the
 // UNDERLYING migration error is returned rather than the dirty flag: that
 // error is the actionable one, and dirty is left set.
+//
+// The bookkeeping being CLEAN is not by itself proof that no interruption
+// happened: a crash inside recovery, between the Force(N-1) above and the
+// re-run committing, leaves exactly that state with N's body already applied.
+// So case 2 is also recognised from a clean start — see forcePastCommittedBody.
 func upWithRecovery(m *migrate.Migrate) error {
 	err := m.Up()
 	if err == nil || errors.Is(err, migrate.ErrNoChange) {
 		return nil
 	}
 	var dirty migrate.ErrDirty
-	if !errors.As(err, &dirty) {
-		return err
+	if errors.As(err, &dirty) {
+		return recoverDirty(m, dirty.Version)
 	}
-	log.Warn().Int("version", dirty.Version).
+
+	// Clean bookkeeping, yet the very first migration attempted collided with
+	// effects already present in the schema. That is the tail of a recovery
+	// interrupted before it finished: this boot re-ran N from a clean N-1 and
+	// hit N's own committed body, so it failed outright instead of with
+	// ErrDirty. Without this the repo is dropped for one more boot — the exact
+	// #33 symptom — before the now-dirty flag routes it through recoverDirty.
+	if alreadyApplied(err) {
+		failed, verr := failedVersion(m)
+		if verr != nil {
+			return err
+		}
+		log.Warn().Int("version", failed).
+			Msg("migration collided with its own committed body; completing an interrupted recovery")
+		if ferr := forcePastCommittedBody(m, failed); ferr != nil {
+			return ferr
+		}
+		return nil
+	}
+	return err
+}
+
+// recoverDirty performs the two-step recovery documented on upWithRecovery for
+// a database left dirty at version.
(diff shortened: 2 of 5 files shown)


knomit:
### Code review

PR #35, opened 2026-07-23
fix(repos): adopt the agent branch on a restored/copied home (#32)

### LINKED ISSUE #32: Restored/copied home cannot bootstrap its agent branch, breaking every write path
Found while testing the v4→v5 graph migration in PR #28 (not caused by it).

## What happens

The agent branch name derives from the local SSH key fingerprint (`internal/app/identity.go:96`). When a knomit home is restored onto a different machine — or a repo database is copied — the new instance generates a new `id_ed25519`, so it looks for an agent branch that does not exist in that database.

`ensureBranch` (`internal/repos/builder.go:256`) tries to create it with `CreateBranch(agent, agent)` (`internal/store/branch.go:282`), i.e. seeding the branch from itself. That cannot succeed when the branch is absent, so the branch is never created and **every write path on that repo is broken**, not just index maintenance.

## How it surfaced

Booting the PR #28 binary against a copy of a real v4 database:

```
WRN schema-mismatch rebuild failed
    error="rebuild: head commit: HeadCommit: resolveRef \"agent/mindev.local-ccd72eef\": branch not found: reference not found"
    branch=agent/mindev.local-ccd72eef repo=knomit-kb
```

The server starts and serves reads normally; the branches API does report `index_state: "error"` for that branch, so it is not silent, but there is no path to recovery without manual intervention.

## Repro

1. Copy a repo database to a different `KNOMIT_HOME` (e.g. `sqlite3 "file:$HOME/.knomit/repos/<repo>.db?mode=ro" "VACUUM INTO '$NEW_HOME/repos/<repo>.db'"`).
2. Start knomit with `KNOMIT_HOME=$NEW_HOME` on a spare port, so it generates a fresh `id_ed25519`.
3. The agent branch in the copied database belongs to the old key; the new one is never created.

## Suggested direction

Adopt an agent branch on first use rather than requiring it to pre-exist: seed the new machine's agent branch from the upstream branch (or from an existing agent branch) when it is absent, the same way a fresh clone bootstraps its agent ref from `origin/main` (`internal/store/repo.go:402-407`).

Worth deciding explicitly whether an orphaned agent branch from the previous machine should be retained, adopted, or garbage-collected.

## Related

PR #28 makes the consequences milder for the graph specifically — migration 000014 now converts `deleted` to TEXT before dropping the typed tables, so a failed heal no longer leaves retracted facts reading as live — but the underlying write-path breakage is untouched and is the real bug.

knomit:
Fixes #32.

## Root cause

The agent branch name derives from the local SSH key fingerprint (`app.agentBranch`). When a knomit home is restored onto a different machine — or a repo DB is copied — a fresh `id_ed25519` yields a **different** branch name that does not exist in that database. `ensureBranch` tried to create it with `CreateBranch(agent, agent)` — seeding the branch **from itself** — which can never resolve a source that is absent. The branch was never created and **every write path** on the repo broke (`HeadCommit` → `resolveRef` → *branch not found*), not just index maintenance.

## Fix

`internal/repos/builder.go`: a new `seedSourceForAgentBranch()` picks the seed source for `CreateBranch`:

- **Agent branch present** (normal case): return itself → `CreateBranch` no-ops, behavior unchanged.
- **Absent** (restored/copied home): adopt the repo's current **HEAD branch** — the previous machine's agent branch, which carries all accumulated knowledge. `CreateBranch` already clones git ref + `branch_commits` + `branch_facts`, so the new machine inherits everything. This mirrors how a fresh clone bootstraps its agent ref from `origin/main`.
- **Degenerate** (HEAD detached / points at the missing branch): fall back to the agent branch itself, so `CreateBranch` fails loudly with the same diagnostic as before rather than seeding from a broken source.

After a successful adoption, `ensureBranch` calls `SetDefaultBranch(agentBranch)` to **move HEAD onto the adopted branch**. This is load-bearing, not cosmetic — see below.

`internal/repos/manager.go`: `ensureBranch` now runs **before** `loadOntology` in `openOne`. On a restored home the configured agent branch is absent until `ensureBranch` adopts it, and `loadOntology` reads (and may rewrite) `domains/ontology.yaml` on that branch. Running `loadOntology` first fell back to the default ontology and skipped the preset-refresh on the first boot after a restore (self-correcting only on the next boot); the reorder loads the adopted branch's ontology on the first boot.

## Why HEAD must follow the adoption

**"HEAD is this machine's agent branch" is an invariant.** Every repo-creation path establishes it — `store.InitRepo`, `InitFromRemote`, `initFromEmptyRemote`, plus the runtime `Manager.Create` paths (`initLocal` / `initClone`) which route through them. Nothing else writes HEAD, and no config lets an operator choose a different one (the name derives from the local key fingerprint). A restored/copied home is exactly the case that violates it.

`ensureBranch` therefore calls `alignHeadWithAgentBranch()` after any successful `CreateBranch`: if HEAD isn't the agent branch, repoint it. Two earlier revisions of this PR got this wrong in different ways, both silent:

1. The first left HEAD on the orphan and called it cosmetic. But `seedSourceForAgentBranch` **reads HEAD** as its adoption source, so HEAD stayed pinned to the machine that first created the repo — a home restored, used, and restored **again** adopted the original branch on the second hop and discarded everything the intervening machine wrote.
2. The second moved HEAD only on the boot that adopted. That made the repair un-retryable: once the agent branch exists no further adoption fires, so a single `SetDefaultBranch` failure pinned HEAD wrong permanently and re-armed the same loss.

Repairing the invariant on *any* boot is what makes it self-healing; it fixes a detached HEAD for free. The check runs only after `CreateBranch` reports success, so HEAD is never pointed at a missing ref (which would dangle it and fail the next `OpenRepo`).

The orphan branch itself is still retained, not garbage-collected — only the HEAD pointer moves off it. Push/sync never consult HEAD (they address the agent ref by name), so this has no remote-facing effect.

## Design decisions

The issue explicitly flagged these (recorded in knomit at `kb/decisions/repos/restored-home-agent-branch-adoption`):

1. **Seed from the existing agent branch, not upstream `main`.** For local-only repos (the dominant case, including knomit's own memory) `main` sits at the init commit — seeding from it would discard every accumulated fact.
2. **Retain the orphaned branch, no GC.** It is not this machine's agent branch (nothing writes to it) and is outside the fetch refspec (never pushed), so it is harmless. GC risks deleting the only copy of data.
3. **Keep HEAD on this machine's agent branch, repairing it on any boot.** Required for correctness across repeated restores (above).

## Notes

- **Origin repos**: an adopted branch has no sync watermark, but `reconcileAgentRebase` already self-heals a missing watermark via a MergeBase fallback — no extra code needed.
- **Latent, not addressed here**: `DropBranch` removes a git ref without checking HEAD, so deleting whatever branch HEAD points at would dangle it and fail the next `OpenRepo`. `DropBranch` has no web/MCP/CLI caller today (tests only), so this is unreachable in production; moving HEAD off the orphan removes the most likely way to stumble into it.

## Testing

- `builder_adopt_test.go` reproduces the restored-home scenario end to end (boot under one agent branch → close → reopen under a different one). It **failed before** with issue #32's exact error and **passes after**. It also asserts the adopted branch inherits the prior branch's facts, that HEAD follows the adoption, and — locking in the ordering fix — that the restored instance loads the previous machine's custom ontology rather than the default (verified this assertion fails under the old `loadOntology`-first ordering).
- `TestEnsureBranch_ChainedRestoreAdoptsMostRecentBranch` covers the second hop: three machines in sequence (A → restore B → restore C), asserting C inherits **both** A's and B's facts. It fails without the HEAD repair.
- `TestEnsureBranch_RepairsHeadWhenAgentBranchAlreadyExists` builds the damaged state directly — agent branch present, HEAD still on the orphan — so the repair is exercised with no adoption firing.
- Full `go test ./...` passes, `internal/repos` + `internal/store/...` also under `-race`; `go build ./...` and `go vet` clean.

### DIFF
--- a/internal/repos/builder.go
+++ b/internal/repos/builder.go
@@ -253,7 +253,7 @@ func (b *repoBuilder) initDefaultGit() error {
 // the origin remote record for the default repo.
 func (b *repoBuilder) ensureBranch() {
 	if b.agentBranch != "" {
-		if err := b.svc.Branches().CreateBranch(context.Background(), b.agentBranch, b.agentBranch); err != nil {
+		if err := b.svc.Branches().CreateBranch(context.Background(), b.agentBranch, b.seedSourceForAgentBranch()); err != nil {
 			log.Warn().Err(err).Str("repo", b.name).Msg("branch create/ensure failed")
 		}
 	}
@@ -266,6 +266,41 @@ func (b *repoBuilder) ensureBranch() {
 			log.Warn().Err(err).Msg("failed to seed origin in remotes table")
 		}
 	}
+}
+
+// seedSourceForAgentBranch returns the branch CreateBranch should seed the
+// agent branch from. Normally this is the agent branch itself: it already
+// exists, so CreateBranch no-ops and the source is never read.
+//
+// But when a knomit home is restored onto a different machine — or a repo
+// database is copied — the local SSH key, and therefore the agent branch name
+// (derived from the key fingerprint, see app.agentBranch), differs from the one
+// recorded in the database. The configured agent branch is then ABSENT, and
+// seeding it from itself can never succeed, so the branch was never created and
+// every write path on the repo broke (issue #32).
+//
+// In that case adopt the repo's current HEAD branch — the previous machine's
+// agent branch, which carries all accumulated knowledge — as the seed. This
+// mirrors how a fresh clone bootstraps its agent ref from origin/main. The
+// orphaned previous branch is retained, not garbage-collected: it is not this
+// machine's agent branch, so nothing writes to it and (being outside the fetch
+// refspec) it is never pushed.
+//
+// Falls back to the agent branch itself when HEAD is detached or already points
+// at the (missing) agent branch, so CreateBranch fails loudly with the same
+// diagnostic as before rather than silently seeding from a broken source.
+func (b *repoBuilder) seedSourceForAgentBranch() string {
+	ctx := context.Background()
+	if _, err := b.svc.Branches().HeadCommit(ctx, b.agentBranch); err == nil {
+		return b.agentBranch // present: CreateBranch no-ops, source unused
+	}
+	def, err := b.svc.Branches().DefaultBranch(ctx)
+	if err != nil || def == "" || def == b.agentBranch {
+		return b.agentBranch
+	}
+	log.Info().Str("repo", b.name).Str("agent_branch", b.agentBranch).Str("adopt_from", def).
+		Msg("agent branch absent (restored home / copied db); adopting from HEAD")
+	return def
 }
 
 // setupIndex configures the search index with the embedder and runs an initial
--- a/internal/repos/builder.go
+++ b/internal/repos/builder.go
@@ -62,6 +62,11 @@ type repoBuilder struct {
 	// origin is configured (detected from the remote's symbolic HEAD).
 	// Defaults to "main" for repos with no origin.
 	upstreamMain string
+
+	// adoptedFrom is the branch the agent branch was adopted from on a
+	// restored/copied home, set by seedSourceForAgentBranch. Empty on every
+	// normal boot. ensureBranch uses it to decide whether to move HEAD.
+	adoptedFrom string
 }
 
 // openStore opens the SQLite-backed store and configures credential encryption.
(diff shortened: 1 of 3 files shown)


knomit:
### Review round 3 — three independent passes (design, line-level correctness, test quality)

**No merge blockers.** The design was checked and holds: the "HEAD is this machine's agent branch" invariant is genuinely universal — every persisted-repo creation path establishes it (`InitRepo` `repo.go:170`, `InitFromRemote` `repo.go:408`, `initFromEmptyRemote` `repo.go:544`, plus `store.Replay` `replay.go:82`, and the runtime `initLocal`/`initClone` which route through them). `CloneFrom` is the one path that leaves HEAD at the remote default, and it is only used for a throwaway preview clone.

Also verified sound: `CreateBranch` really does guarantee the ref exists on a nil return, so `alignHeadWithAgentBranch` cannot dangle HEAD; the `openOne` reorder is dependency-free in both directions; `ensureBranch` runs before the commit observer is wired and before `activate()`, so nothing can race it; and the watermark reasoning in the PR body checks out.

**What changed in this round (20e4ae15, tests only):** two mutations passed the previous tests, meaning two documented decisions were unenforced.

| Mutation | Before | Now |
|---|---|---|
| GC the orphan branch after moving HEAD off it | all tests passed | fails — orphan retention asserted |
| Repair HEAD regardless of `CreateBranch` success | all tests passed | fails — new detached-HEAD test |

Also: facts are now written with parseable frontmatter (the previous bodies had none, so they were never indexed and `CreateBranch`'s `branch_facts` clone — the mechanism adoption depends on — went untested); `Deep` `Verify` is asserted clean on the adopted branch; the write path is now the *first* assertion in the adoption test (on unfixed code it used to report an ontology mismatch instead); and the `openOne` ordering assertion is split into its own test, since it was the sole guard of that commit while buried under a name about adoption.

**Filed as follow-ups rather than expanding this PR:**

- #36 — a **stale** agent branch is never reconciled, and HEAD is repointed at it with no ancestry check. Reproduced: A → B → back to A leaves A on its stale branch, silently unable to see B's work, and a later C inherits A's lineage. Verified byte-identical before and after this PR, so not a regression — but this PR converts a loud failure into a silent one, which is worth tracking. Note the branch name is `agent/<hostname>-<fp8>`, so a **hostname flip alone** (macOS `.local`/`.lan`) triggers it — both #32 and this PR describe it as key-derived only.
- #37 — adoption does not rewrite the origin fetch refspec for non-default repos.
- #38 — coverage for origin repos, secondary repos, and the `Manager.Restore` path (all verified working today).

PR #39, opened 2026-07-24
fix(repos): take over a repo by recorded owner on agent-branch change (supersedes #35, fixes #32)

### LINKED ISSUE #32: Restored/copied home cannot bootstrap its agent branch, breaking every write path

knomit:
Supersedes and reverts #35. Fixes #32.

#35 fixed #32 by seeding the new agent branch from the repo's **HEAD** and adding HEAD-repair machinery to keep that seed source trustworthy. That was wrong for a multi-agent repo: several agents share a repo through its origin and their branches are fetched into `refs/heads`, so HEAD does not identify the branch *this* database writes to. Seeding from HEAD can fork an instance off **another agent's lineage**. This PR reverts #35 and fixes #32 correctly.

## Commits

1. **Revert #35** — undo the HEAD-based adoption and its repair machinery.
2. **Take over by recorded owner (#32)** — the real fix.
3. **Recover the upstream branch on reboot** — an unrelated pre-existing bug found in the same code.

## The fix (commit 2)

The agent branch name is `agent/<hostname>-<key-fingerprint>` (`app.agentBranch`), so it moves if the SSH key is regenerated or the machine is renamed. When it moves, the configured branch is absent and every write path breaks (`HeadCommit` → `resolveRef` → *branch not found*) — issue #32.

Record which agent branch owns each repo database in a `meta` row (`agent_branch_owner`). When this instance's derived name does not match the record, **take over**: seed the new branch from the recorded owner (which carries the accumulated knowledge) and claim ownership.

- **Seed from the recorded owner, never from HEAD.** The recorded owner is the only branch known to belong to this database. `TestEnsureBranch_DoesNotTakeOverAnotherAgentsBranch` parks HEAD on a foreign agent's branch and asserts the takeover ignores it — this test **fails against #35's logic** and passes here.
- **No usable record → fail loudly.** `CreateBranch` errors rather than seeding from a guess; origin is the source of truth and re-cloning rebuilds the database (`TestEnsureBranch_NoRecordedOwnerFailsLoudly`, which also asserts the repo still opens afterward).
- **HEAD is no longer load-bearing.** All of #35's `alignHeadWithAgentBranch` / chained-restore HEAD machinery is gone — nothing in knomit reads local HEAD once the seed source stops depending on it.
- **Ordering kept:** `ensureBranch` runs before `loadOntology` in `openOne`, so the adopted branch exists before the ontology is read off it.

## Commit 3 (unrelated, pre-existing)

`upstreamMain` is resolved only by `initDefaultGit`, which runs on the first boot alone. Every later boot left it empty, so the fallback to the literal `"main"` silently rewrote a `master`-tracking repo's stored upstream and fetch refspec on the second boot, and aimed the startup index sync at a branch that does not exist. `openGit` now rehydrates it from the stored remote record. Regression test boots twice against a `master`-default remote and fails without the fix (`"main"` vs `"master"`).

## Testing

- `go build ./...`, `go vet ./...`, `go test ./...` all clean.
- New/rewritten `internal/repos` tests: restored-home takeover, chained restore, foreign-agent rejection, no-owner-fails-loudly, ensureBranch-before-loadOntology ordering, and the upstream-reboot regression.


knomit:
### Code review

PR #41, opened 2026-07-25
feat(okf): knomit-okf — publish a knomit KB as a conformant OKF v0.2 repository

knomit:
## knomit-okf — publish a knowledge base as an OKF repository

Turns any knomit knowledge base into a publishable, syncable **Open Knowledge Format (OKF) v0.2** repository, using nothing but a git URL:

```bash
knomit-okf clone https://github.com/knomit/knomit-kb foobar
cd foobar
git remote add origin git@github.com:me/my-kb-okf.git && git push -u origin main

# later — by you, or by anyone who cloned your published repo
knomit-okf sync
git push
```

`foobar` **is** the OKF export. No second checkout, no cache directory, and no state held by the knomit server.

> **This PR changed shape.** It originally generated bundles *inside the server* and served them as `refs/heads/okf/<branch>` plus a tarball route. Review found that approach carried a Critical regression; the export is now a standalone tool and the server-side machinery is deleted. The valuable half — the pure mapper — is kept unchanged. See [Why this replaced the server-side export](#why-this-replaced-the-server-side-export).

Export-only in v1; the mapping is kept round-trippable so an OKF→knomit importer is a later, non-breaking addition.

### Verified against live corpora

| repo / branch | facts | bundle files | clone | conformance |
|---|---:|---:|---:|---|
| `knomit-kb` `main` | 297 | 910 | — | PASS |
| `knomit-kb` `agent/mindev…` | 328 | 1012 | — | PASS |
| `cyberai-kb` `main` | 1113 | 2816 | **18.0s** | PASS |
| `cyberai-kb` `agent/backup` | 787 | 1969 | 8.6s | PASS |

0 skips, 0 warnings. Two *independent* clones of the same source produce the **identical output commit SHA** (`0e3eadf4`) — export commits are timestamped from the source commit, never the clock. A re-`sync` with no upstream change makes no commit at all. Total `.git` for source history plus bundle: **6.3 MB**.

---

## Why this replaced the server-side export

Holding state caused the problems — nearly all of them downstream of it:

| Problem | Cause | Now |
|---|---|---|
| **`knomit verify` broke** (0 → ~1000 errors on a real corpus) | generated refs lived in `refs/heads/*`, so every ref enumerator saw them as branches | **gone** — no refs are generated |
| Unbounded store growth | one snapshot commit per regeneration, never pruned | **gone** — nothing is stored |
| 3.5 s per branch inside unauthenticated `/info/refs`, no singleflight | generation on the fetch path | **gone** — generation is user-initiated |
| Non-fast-forward races | unlocked read-modify-write on the ref | **gone** — no shared ref |
| Marker key could not capture all its inputs | hand-maintained `MapperVersion` | **gone** — no cache to invalidate |
| Push guard panicked in a bare goroutine, killing the server | defensive guard against leaking `okf/*` | **gone** — nothing to leak |

Six commits sat on `okf/main` in the live store, **all of the same source commit** — they recorded mapper releases, not knowledge changes. Now a commit exists only when the exported content differs, because git computes the diff and the mapper is deterministic.

**Deleted:** `EnsureOKF`, `okfWriteTree`, the ref write, the marker, `okf_trigger.go`, the `/okf/{branch}.tar.gz` route, `push_guard.go`, `store/git/okf_marker.go`, and `okf.MapperVersion`.

---

## Architecture

Three units, each independently testable:

- **`internal/okf`** — the pure mapper. No git, no SQL, no HTTP, no clock. Facts in, in-memory bundle out; also provides `Validate`. **Kept unchanged** apart from defect fixes.
- **`internal/okfsource`** *(new)* — reads a KB from **any go-git storer** and produces the mapper's inputs. A lift of the read half of `internal/store/okf.go`, which touched no SQL. This is what lets the exporter run against a plain `git clone` with no knomit server involved.
- **`tools/okf`** — the CLI: git plumbing and the two commands.

The tool must live in the knomit module: Go forbids importing `knomit/internal/...` from outside it.

### Repository layout

One directory. The source history lives in the same `.git`, under a private ref namespace, and is **never checked out**:

```
foobar/
├── .git/
│   ├── config                     remote "knomit-source" → the KB URL
│   └── refs/knomit-okf/source/*   fetched source branches (not checked out)
├── .knomit-okf.yaml               sync metadata
├── index.md   log.md   kb/   views/     the bundle — the whole working tree
│                                   (kb/ carries per-folder index.md and log.md)
```

Because `refs/knomit-okf/source/*` sits outside `refs/heads/*`, git's default push refspec **cannot** match it — so a *private* KB can be the source of a *public* export. Verified by a test that pushes to a bare remote and asserts only `refs/heads/*` arrives. The one leak path, documented and deliberately undefended in code: `git push --mirror`.

Output branches **mirror source branch names**; a second source branch becomes its own **orphan** output branch, since the bundles are unrelated snapshots rather than descendants.

### The owned-paths boundary

The tool owns exactly `index.md`, `log.md`, `kb/`, `views/`, `.knomit-okf.yaml` — writing them completely and deleting anything under them it did not write. Both halves are load-bearing in opposite directions:

- **Deleting is mandatory.** Overlaying files can never *remove* them, so a retired fact's document would stay published forever, contradicting the `views/retired.md` in the same bundle.
- **Deleting anything else is forbidden.** A publisher's `README.md`, `LICENSE`, and `.github/` must survive a sync. The naive "clean the working tree" destroys their work, so `reconcile` *rejects* an unowned path rather than writing it, and only owned paths are staged — a publisher's uncommitted edits are never swept into an okf commit.

---

## Bundle shape

```
index.md          okf_version: "0.2"
log.md            root changelog: creations, plus events with no folder left
kb/               authored ontology, structure preserved verbatim
  **/index.md     per-directory index
  **/log.md       that folder's changes — revisions and withdrawals
views/            derived: domains/, entities/, synthesis, hypotheses,
                  methodology, retired
```

`kb/` and `views/` are kept apart so a knomit topic legitimately named `domains` can never collide with a generated view. A bundle path is identical to the fact's `knomit_path`, so internal citations resolve without prefix juggling.

### What a consumer gets

- **A traversable knowledge graph.** Citations resolve to links labelled with the cited fact's title, and every fact also names **what cites it** — knomit's `DERIVED_FROM` graph, walkable in both directions.
- **Progressive disclosure that works.** Every directory index links its children *and* its concept documents, carrying the ontology's own authored descriptions (`principles` → "Designer-authored intent — mission, philosophy, anti-patterns, UX taste").
- **Per-fact history** — how a belief evolved: `2026-04-27 · update — confidence 0.79 → 0.82`. Rendered only when a fact actually changed, where "changed" is one shared rule (`MeaningfulRevisions`) that `log.md` uses too — a ref *count* moving is not a change, since `RefCount` is an int and can name neither which ref nor why.
- **What the base withdrew.** `views/retired.md` lists every retracted or superseded fact with its successor linked. Most knowledge bases claim to be current; few can show what they corrected.
- **Cross-cutting views** OKF defines no structure for: by domain, by entity, and chronological digests of the derived types. Long indexes get an alphabetical jump bar; digests group by month then day.
- **Links are relative**, because GitHub resolves a leading `/` against the repo root and publishing there is the intended distribution path.

### Where knomit's model shows

| OKF v0.2 | knomit source |
|---|---|
| `sources[]` | refs that resolve to a followable artifact |
| `generated: {by, at}` | `origin` + authoring commit |
| `status` | confidence bands, hypothesis type |
| `knomit_*` | the lossless round-trip channel |

Four deliberate refusals, because over-claiming corrupts the signal consumers rely on:

- **`verified` is never emitted.** knomit records no verification events; synthesising them from revision commits would fabricate exactly what trust tiers are derived from.
- **`generated.by` never claims `human:`.** knomit's `authored` means hand-written by a human *or* an agent, indistinguishable at export.
- **`src://` anchors stay inert.** A KB repo is usually a different repo from the code it documents, so deriving links from the bundle's own remote would emit plausible links to paths that don't exist. `RenderOpts.ResolveSource` is the seam.
- **Retired facts are an index, never documents.** Their claims were disavowed, and a conformant consumer may ignore `status: deprecated` — so an ingestible document would invite re-ingestion of withdrawn knowledge.

---

## Commands

```
knomit-okf clone    [-b <branch>] [--publish-source] [auth] <kb-url> <dir>
knomit-okf sync     [-b <branch>] [--source <url>] [--publish-source] [auth]
knomit-okf branches [--source <url>] [--no-fetch] [auth]

auth flags (all optional; a knomit server needs none):
  --token <t> | --token-file <p>   access token   ($KNOMIT_OKF_TOKEN)
  --username <u>                   basic-auth user for the token, default "git"
  --ssh-key <path>                 SSH private key ($KNOMIT_OKF_SSH_KEY)
                                   passphrase via $KNOMIT_OKF_SSH_PASSPHRASE
```

`sync` skips rendering entirely when the fetched head already equals `synced_commit`, the recorded release matches this build, and the owned paths are clean. It **never pushes** — your remote, your credentials, your cadence.

`sync -b <other>` exports a second source branch onto an output branch of the same name, created as an orphan. `branches` exists because those names were otherwise undiscoverable — the only way to find one was to pass a wrong `-b` and read the error:

```
  BRANCH                                      EXPORTED  STATUS
    agent/Alexs-MacBook-Air-6.local-60def18b  —         not exported
  * agent/backup                              6ad89f79  up to date
    agent/debian-dev-24971f37                 —         not exported
    main                                      65788b13  up to date
```

It reads each output branch's `.knomit-okf.yaml` from its **committed tree**, not the working tree — only one branch is checked out, so reading from disk would make every other branch look unexported. Drift reports as `3 commits behind`; a source that moved incompatibly reads `diverged` rather than a misleading count.

Fetch **prunes**: `refs/knomit-okf/source/*` mirrors upstream branches exactly as remote-tracking refs do, and without pruning a branch deleted upstream would linger forever — `branches` would report it live and `sync -b` would export a branch the knowledge base no longer has.

### Progress output

Every stage announces itself **before** the work runs. A silent multi-minute clone is indistinguishable from a hang, which is exactly how the performance bug below was first reported.

```
knomit-okf 0.5.0.78233d95

  ✓ Fetching    5 branches                               307ms
  ✓ Reading     1113 facts · 1656 events · 226 retired     9.8s
  ✓ Rendering   2816 documents                             41ms
  ✓ Validating  conformant with OKF 0.2                    35ms
  ✓ Writing     2817 changed · 0 removed  (of 2817)       246ms
  ✓ Staging     2817 files                                 7.4s
  ✓ Committing  9f83d523 on main                          164ms

✓ Cloned main into cyberai-kb in 18.0s
```

A later `sync` does proportional work, and says so — this is the same 3038-document bundle after 8 facts were added upstream:

```
  ✓ Reading     1208 facts · 1778 events · 247 retired    11.6s
  ✓ Rendering   3037 documents                             41ms
  ✓ Writing     44 changed · 0 removed  (of 3038)         178ms
  ✓ Staging     44 files                                  189ms
  ✓ Committing  f517b81b on agent/mindev.local-8ef0cd32    14ms
```

On a terminal the open stage updates in place with live counters (throttled to 60ms). Off one — a pipe, a CI log, a redirect — it degrades to exactly one plain line per stage: no ANSI, no carriage returns. `NO_COLOR` and `TERM=dumb` are honoured. Both modes are covered by tests, since silent rot here is invisible until someone reads a log.

Source URL precedence: `--source` → the `knomit-source` git remote (local, never travels) → `source:` in `.knomit-okf.yaml`. That last field is written **only** with `--publish-source`, so a private KB's address is never published by default; it exists so a stranger who cloned a *published* repo can sync it. An already-published URL survives a bare `sync`, and carries to an output branch created later by `sync -b` — silently un-publishing it, on this branch or the next one, would break every downstream clone.

`--source` is genuinely one-off: fetching goes through an anonymous remote and never touches stored config, so an override cannot silently repoint the repository. **`clone` is the only command that records where knowledge comes from**; a KB that has genuinely moved is repointed deliberately with `git remote set-url knomit-source <new>`.

### `.knomit-okf.yaml`

```yaml
# maintained by knomit-okf — https://github.com/knomit/knomit
branch: main
synced_commit: 8f238008beea3af8c106ace370cec86cc852480a
tool_version: 0.5.0.fd405925
```

Conformance-safe: OKF's rules constrain only `.md` files. `synced_commit` also makes each sync's diff self-describing — the source commit moves alongside the content it produced. `tool_version` replaces the deleted `MapperVersion`: recording which tool produced a bundle belongs in the artifact, not in a Go constant. It is also *read* — its `Major.Minor.Patch` part gates the no-op skip, so a new release re-exports once per branch while a rebuild of the same release does not.

---

## Authentication

A knomit server's git endpoint is unauthenticated, so the common case needs nothing. But a KB hosted on **GitHub, GitLab, or Bitbucket** needs credentials, and go-git supplies none by default — SSH fell back to ssh-agent only, and HTTPS was anonymous. That made a private KB reachable only via a `git clone --mirror` dance, which the docs previously recommended and this now deletes.

Two optional mechanisms, resolved per run from flags and environment:

```bash
knomit-okf clone --token $GH_TOKEN https://github.com/me/private-kb my-kb
knomit-okf clone --username x-token-auth --token $BB_TOKEN https://bitbucket.org/me/kb my-kb
knomit-okf clone --ssh-key ~/.ssh/deploy_key git@github.com:me/private-kb.git my-kb
```

Pure go-git — **no git CLI dependency, no credential helpers, no `.netrc`, no `~/.ssh/config` parsing, and no new dependencies.** `authFor` returns a nil `AuthMethod` for local paths and anonymous HTTP, so the existing knomit-server path is bit-for-bit unchanged and is pinned by test.

**The token is the basic-auth *password*, never a Bearer header.** go-git's `TokenAuth` sends `Authorization: Bearer`, which all three hosts reject for git-over-HTTPS. A dedicated regression test pins this. The username field is not uniform — GitHub and GitLab ignore it, Bitbucket access tokens require the literal `x-token-auth` — hence `--username`, which exists specifically so Bitbucket users aren't pushed onto the URL form that persists a token to `.git/config`.

**SSH resolves `--ssh-key` → ssh-agent → `~/.ssh/id_ed25519`/`id_rsa`/`id_ecdsa`.** An explicit key that won't load is a hard error, never a silent fallback. The agent is *probed for signers*, because `NewSSHAgentAuth` succeeds whenever an agent **socket** exists even with zero identities loaded — a very common state, and without the probe the default-identity fallback would be unreachable dead code. `known_hosts` verification is never disabled; since go-git offers no interactive "continue connecting (yes/no)?" prompt, a first contact with an unknown host now fails with the `ssh-keyscan` command that fixes it — and a host key that *mismatches* an existing entry is reported as a different failure that never gets that advice (see the fourth review pass).

The SSH passphrase is environment-only (`$KNOMIT_OKF_SSH_PASSPHRASE`) — a passphrase on argv is visible to every user via `ps`.

### Credentials never reach disk or output

Two leaks were found and closed, one of them pre-existing:

| Leak | Where | Now |
|---|---|---|
| `--publish-source` wrote the source URL verbatim into the **committed** `.knomit-okf.yaml` | `renderFiles` | Redacted before writing, with a note when it strips anything |
| The raw URL was interpolated into every fetch error, so a token printed to the terminal | `fetchSource`, `remoteHeadBranch` | Redacted at construction; wrapped error text scrubbed too |

The second was subtler than it looks. Every command already redacted the URL in its *progress* line, but not in the error one call below — and fixing that alone was insufficient: Go's own `*url.Error` strips only the **password**, so a token riding as the *username* still printed in full, and `net/url.Parse` errors quote their entire input verbatim. Both reach the terminal through `%w`. The fix scrubs wrapped error text while keeping the `errors.Is` chain intact, which the 401 hint depends on.

Redaction is deliberately asymmetric: for `http(s)` the whole userinfo goes, because a token may be *either* field; for `ssh` only the password goes, since the username is a login name and stripping it would publish a source URL that no longer resolves.

**Known limitation, not addressed here:** a credential embedded in the URL passed to `clone` is still stored unredacted in the local `.git/config` as the `knomit-source` remote. That is pre-existing behaviour and the file is never pushed, but it is on disk — so the docs now steer to `--token` and say so plainly. A token in a *query string* (`?access_token=…`) is likewise still printed; it is not a mechanism any of the three hosts use for git-over-HTTPS.

### Test coverage

An end-to-end test clones against a real smart-HTTP server (`net/http/cgi` wrapping `git-http-backend`) gated on basic auth: anonymous is rejected, `--token` succeeds, and the token appears in neither stdout nor the error. The zero-identity-agent path is covered by serving a real empty `agent.Keyring` over a unix socket. Both were verified non-vacuous by reverting the implementation and confirming they fail.

---

## Packaging

`knomit-okf` ships everywhere `knomit-bridge` does — the server tarball, the macOS `.app`, and the Linux desktop tarball — built with `VERSION_LDFLAGS` so `tool_version` is populated. The desktop app symlinks it to `<home>/bin` on launch, as it already does for the bridge: a CLI reachable only at `Knomit.app/Contents/MacOS/knomit-okf` is one nobody runs.

---

## Notable bugs found and fixed under test

- **`ParseConcept` did not round-trip the body** — `stripCitations` cut only at `# Citations`, but `# Related` is emitted *before* it, so that section leaked into every reconstructed body. The existing test never asserted `Body`, which is why it passed.
- A tree fold-order bug producing zero subtree hashes and non-deterministic SHAs.
- go-git's canonical tree-entry ordering (`ErrEntriesNotSorted`).
- `business` rendering as `busines` — a strip-trailing-`s` singularization fallback, removed.
- Same-second revisions rendering a **backwards** delta (`0.9 → 0.5`) via a tie-break carrying no chronology.
- **Merges counted as revisions** — first-parent-only diffing reported every file a merge carried in as modified. 55% of recorded history on `core`.
- The same first-parent blind spot misclassifying retirements.
- `subsumed by distilled fact` classified as *retracted* rather than *superseded* — a claim absorbed into a synthesis is not a claim withdrawn as wrong.
- **Guards that didn't guard.** Several tests were verified by deliberate breakage rather than trusted. One is worth calling out: a sort-stability test could never have failed. The common wisdom — "Go's `sort.Slice` is insertion sort below n≈13, so use a bigger fixture" — is only half the story; above that threshold pdqsort *still* short-circuits on an already-ordered input, so a 20-element all-equal fixture passes identically under `sort.Slice` and `sort.SliceStable`. Rebuilt around the shape the sort actually exists for (commit-walk order, ties within groups) and confirmed red under the mutation.
- **Staging was quadratic.** `wt.Add` recomputes the *entire* worktree `Status()` per call, so staging a bundle cost `files × O(files)`: **115.9s** for 1969 files, against **3.6s** for a byte-identical index via `AddWithOptions{SkipStatus: true}`. Invisible at test scale — a 10-file fixture stages instantly — and it presented as a hang on a real corpus, not as slowness. `SkipStatus` is safe here only because `reconcile` has just written every path being staged: no directories, nothing missing, which are the two cases `Status` exists to resolve. Deletions are handled by pruning index entries directly.
- **`--source` was not one-off, despite saying so.** Its help read *"override the KB URL for this run"*, but the fetch kept the stored `knomit-source` remote in step with whatever URL it was given — so a one-off silently repointed the repository, and every later bare `sync` followed the wrong knowledge base. Pruning compounded it into data loss: one `sync --source <other-kb>` would also drop the original KB's source refs, since the fetch prunes whatever the new remote does not advertise. Two individually reasonable changes combining into something destructive.
- **"N commits behind" overcounted across merges.** The walk halted *at* the exported commit rather than excluding its ancestors, so from a merge it descended into the second parent and counted commits that are ancestors of the stop commit by another path. On `A←B`, `A←C`, `M(B,C)` with stop `B` it reported 3 instead of 2.
- **A sync rewrote and re-staged the whole bundle, and misreported it.** Reported by a user, from output alone: a sync that changed 44 documents printed `3038 written` and re-staged all 3038, which reads as "everything was regenerated" — 15.2s of a 27.2s run spent re-adding identical files. The *commit* was always a correct 44-file diff, so no test caught it; every test passed while the tool described its own behaviour falsely. `reconcile` now skips files already identical on disk and returns what it wrote; staging covers those plus anything whose index entry hash differs from the rendered blob. Measured on that bundle: staging **15.2s → 189ms**, the run **27.2s → 12.2s**, with a tree byte-identical to a full re-stage (verified against a fresh clone, and pinned by a test that fails when the skip is made unsound).

### Second review pass

Five findings, all of the same family — the tool knew something and did not say so, or refused without saying why. Each is pinned by a test verified against the behaviour it replaced.

- **A truncated history walk corrupted dates in silence.** `okfHistoryResult.Warnings` was declared, threaded through `Snapshot`, and rendered by `renderFiles` — but never assigned, so the 5000-commit bound reported nothing. Past it a fact whose every revision fell off the walk has no authoring time, and `okfReadFacts` stamps it with the export commit's: a 2024 fact publishes today's date as `timestamp` and `generated.at`. Retirements older than the bound vanish from `views/retired.md`, the one view whose whole claim is showing what the base withdrew. Worst, it is not stable — commits accrue at the tip and push older ones off the back, so a date already published can **change** on a later sync. The bound now reports itself, and the facts that fell back to the export date are counted separately: the bound is the cause, that count is the damage, and only the second tells a reader how much to distrust. The verified corpora sit well under 5000, which is why nothing surfaced it.
- **A fact that failed to parse was dropped without a word.** Right for a stray `README.md` under `kb/`, wrong for a corrupted fact — that is knowledge silently deleted from a published base. `looksLikeFact` separates them on either of two signals: a uuid8 basename (how every fact knomit writes is named, true even of a file mangled beyond recognition) or a frontmatter fence (shaped like a fact, something inside it broken). Ordinary markdown has neither and stays silent, so the warning that matters is not trained away.
- **Any uncommitted publisher edit blocked `sync -b`.** `wt.Checkout` was called with neither `Force` nor `Keep`, which go-git routes through `Reset(MergeReset)` — refused whenever any tracked file differs from the index, and reported as a bare `worktree contains unstaged changes` naming nothing. So the design intent inverted: *"a publisher's uncommitted edits are never swept into an okf commit"* became *"a publisher's uncommitted edits block the sync"*. `TestCLI_PublisherReadmeSurvivesSync` **commits** its README, which is why nothing caught it. `Keep` is not the fix — it maps to `SoftReset`, so the switch would silently not happen — and `Force` is not either, since a hard reset is entitled to files outside the owned paths. The two causes are now split: a dirty file the **publisher** owns is refused, naming the files and the way out; a dirty file **we** own is restored from the index and the switch proceeds, because export is about to rewrite every owned path anyway. Untracked files are neither — go-git carries them across a switch.
- **A failed `clone` left a directory that blocked the retry.** The directory was created and a repository initialised before the first thing that can plausibly fail, so any failure returned with `.git/` on disk — and `ensureEmptyDir` refuses a non-empty target, so the obvious retry died with `<dir> is not empty` and demanded a manual `rm -rf`. Newly reachable, too: auth landed in this branch, and a rejected token is the first thing many users will meet. Cleanup distinguishes what we made from what we were given — a directory this run created is removed; one the **user** created is emptied back to the state `ensureEmptyDir` accepted, never deleted. The same trap as `--source` above: the unconditional, obviously-correct-looking operation is the destructive one.
- **A credential flag aimed at the wrong transport was dropped on the floor.** `--ssh-key` with an `https://` source fetched anonymously and then failed with a 401 whose hint said *"pass `--token`"*; `--token` with an `ssh://` source fell through to whatever ssh-agent happened to hold. Both send the user hunting for a credential problem that is really a command-line problem. `authOpts.specified()` already existed for exactly this reporting but was wired only into `branches --no-fetch`. An unusable flag is now an error, for the same reason an unloadable `--ssh-key` already was. **Only flags** — `$KNOMIT_OKF_TOKEN` and `$KNOMIT_OKF_SSH_KEY` are ambient CI configuration, not something the user asked this run to do, so an https clone in a shell that exports an ssh key keeps working.

### Third review pass

Six findings. Two changed behaviour, four removed a way to be misled. Each behavioural fix is pinned by a test verified against the code it replaced.

- **A published source URL did not reach a branch created by `sync -b`.** The orphan path blanked the config wholesale, so the new branch committed a `.knomit-okf.yaml` with no `source:`. The `knomit-source` remote is local-only and never travels — on a stranger's clone that field is the **only** route back to the knowledge base, so the branch was stranded and a bare `sync` there died with `no knowledge-base URL`. `Branch` and `SyncedCommit` still reset, because they describe the branch we just left; `Source` does not, because it describes where the *repository* gets its knowledge. Carrying it publishes nothing new — it is already committed on a sibling branch — and an unpublished source stays empty on every branch, which is pinned separately.
- **`tool_version` was written and read by nothing.** The no-op skip was keyed on `synced_commit` alone, so a mapper improvement never reached a knowledge base that had stopped moving: the bundle stayed at the old mapper's output until the source happened to change, which may be never. Exactly the staleness this PR indicts in the deleted `MapperVersion` marker — except the input *was* captured, just never compared. The naive fix compares `version.String()` and is wrong: it embeds the build SHA, so every developer rebuild would re-render, rewrite `tool_version`, and commit a bundle whose content nobody changed. It compares the **release**, so the residue is one re-export per release per branch — a true record that the bundle was regenerated and re-validated under it.
- **Two output branches claiming one source collapsed silently.** The join was a bare map write, so the winner depended on ref-iteration order and the table could say something different on each run. It happens after `git branch -c` of an output branch, which copies its config along with everything else. Both claimants are now named and the row resolved deterministically. Reported, not resolved: `sync -b` updates only one of them, and which bundle is authoritative is the user's call.
- **The `branches` table mixed byte and rune widths** — sized with `len(r.source)`, padded with `len([]rune(s))` — so a multi-byte branch name over-padded. The width was also computed over the source name alone while the cell can print `source → output`, pushing `STATUS` out of line on exactly the rows worth reading.
- **`writeConfig` bypassed the owned-paths pass.** Used only by its own test, and it wrote the config straight to disk — around the reconcile-and-stage pass that `marshalConfig` exists to feed, and that the owned-paths boundary depends on. A convenience writer sitting next to `marshalConfig` is the obvious thing for a future caller to reach for, so there is no longer one to reach for.
- **`fileHasContent` returned an error it could not produce.** Both paths already returned `(false, nil)`, so `reconcile` carried a check that could never fire. The swallowing is correct and stays — a missing or unreadable file is simply "not the same", and this must never be why a file is skipped *or* why a reconcile fails. The signature no longer advertises a contract it does not have.

### Fourth review pass

Ten findings. Five destroyed or lost something silently, one gave dangerous security advice, and four removed a way to be misled. Every behavioural fix is pinned by a test confirmed failing against the code it replaces — ten mutations, each reverting one fix, each caught.

- **A failed `sync -b <new>` stranded the repository and deleted the bundle it had.** Preparing an orphan branch moves `HEAD` and empties the index *before* the export that justifies it has produced a byte, and it also `RemoveAll`'d every owned path. So a render failure — an unreachable object, a non-conformant bundle, a bad token — left the user on an unborn branch, with an empty index, with the previous bundle gone from disk, and with nothing saying `git switch <old>` would fix it. Exactly the trap this branch already fixed for `clone`: the unconditional, obviously-correct-looking operation is the destructive one. The preparation now returns a rollback that restores `HEAD` and the index, and the run says `nothing was committed`. The `RemoveAll` is gone outright — `reconcile`'s prune produces an identical tree, so it bought nothing and was the only step that destroyed on-disk content up front.
- **The two routes to an unborn branch disagreed.** "`HEAD` already points at it" returned `created` *without* clearing the index — reachable after an interrupted run, or a `git checkout --orphan` — so whatever the previous branch had staged would be committed as the new branch's first commit, silently defeating the orphan's whole point. Both routes now share one preparation.
- **`sync` and `branches` did not work from a subdirectory.** `PlainOpen` does not search upward, so both died from anywhere but the root with *"run knomit-okf clone first"* — advice that is wrong for precisely the user who did. They now search as git does, and resolve owned paths against the **worktree root** rather than the cwd, so a `sync` from inside `kb/` cannot write a second bundle into a subdirectory. Searching upward newly makes a stray `sync --source` inside an *unrelated* checkout reach a repository, so one that has neither the `knomit-source` remote nor a config is refused rather than rendered into.
- **An owned root that is a symbolic link wrote outside the repository.** `ln -s /etc kb` and `reconcile`'s write loop lands in `/etc`, after which the prune removes the link and orphans what it wrote — the owned-paths boundary held on paths while the filesystem walked straight through it. git cannot represent a symlinked directory holding tracked files, so this only ever comes from a local hand-edit; it is now refused before anything is written.
- **A host-key MISMATCH was told to run `ssh-keyscan`.** The hint matched the substring `knownhosts`, which catches *both* failures — and they are fixed by opposite actions. An unknown host wants `ssh-keyscan <host> >> ~/.ssh/known_hosts`; a key that does **not match** an existing entry is a re-keyed server or a machine-in-the-middle, and appending it silences the one check standing between the user and an interception. Classification now prefers the typed `*knownhosts.KeyError`, whose `Want` field separates them exactly. The substring fallback stays, because the typed error does not survive every transport's wrapping and losing the hint on a dependency update would be worse than a fuzzy match — but it defaults to the *safe* reading, and is kept narrow enough that `ssh: no common algorithm for host key` is not dressed up as an interception. A mismatch gets `ssh-keygen -R <host>` and an explicit "do NOT run ssh-keyscan".
- **`--username` without a token was dropped on the floor.** It names the basic-auth field a *token* rides in, so alone it authenticates nothing: the fetch went out anonymously and failed with a 401, sending the user after a credential problem that is really a command-line one. The same family as `--ssh-key` against an https source, fixed the same way — the third pass wired that rule for two flags and missed the third.
- **A tag named `index` lost its hub page.** It slugifies to `index.md`, which `renderHubs` writes the *directory* index to afterwards, so the hub was overwritten and every concept document in that group linked to the listing instead. Silent — no skip, no warning. The name is now reserved.
- **`ParseConcept` truncated a body at a heading that only looked generated.** `stripGenerated` cut at the first `# Related`/`# Citations`/`# Cited by`/`# History` found *anywhere*, so an authored section named like a generated one — or one quoted in a fenced code block, which a fact *about this exporter* would naturally contain — silently lost everything below it. The same shape as the `stripCitations` defect found earlier in this PR, one layer further out. It now peels the sections off the **end**, accepting a candidate only when nothing below it outranks a generated section. The one case no rule can separate is documented rather than papered over.
- **A repeated ref was presented twice.** A fact naming the same ref twice emitted two identical `sources[]` entries and two Citations lines. Deduplicated for the two places a ref is *presented*, keeping authored order — the existing `dedupe` sorts, which would have reordered a reader's citations. `knomit_refs` stays verbatim: it is the lossless round-trip channel, so a repeat there is data rather than noise.
- **Two counts that could run long or read wrong.** `countNewCommits` was the one unbounded walk left, with no progress stage, run once per exported branch; the counting walk is now bounded and reports `N+ commits behind` when it stops — the exclusion set stays complete, which is what keeps `N` a true lower bound rather than an over-count — and `branches` announces the comparison. Separately, `renderFiles` printed one line per skipped fact and never the total, so a corpus with hundreds of them buried the headline; the count is stated first and the reasons capped, matching `okfsource`'s own degradation reports. Also removed a hand-rolled `itoa` that returned `""` for a negative input.

### Fifth review pass — what a changelog is for

`log.md` and the per-fact `# History` each decided independently what was worth reporting, and both decided badly. Measured on a live 1208-fact corpus:

| | before | after |
|---|---:|---:|
| `log.md` (root, before the split below) | 1778 rows, **176850 B** | 319 rows, **35755 B** |
| — Creation rows | 1456 | 59 |
| — Update rows | 322 | 260 |
| facts carrying `# History` | 136 | 72 |
| delta lines saying only `refs updated` | 79 | 0 |
| facts reported added | 1456 | 1456 |

- **Two views, one rule.** `MeaningfulRevisions` is now the single definition of "worth telling a reader about", shared by both, so they cannot make different claims about the same commit. `okfsource` stamps each log event with the delta of the revision it was recorded *with*, rather than re-deriving it by matching on timestamps — the one key that is not unique here, since a batch commit stamps every path it touched with the same second.
- **A ref count alone is not a change.** It was 79 of 162 delta lines and the entire content of 64 of the 136 `# History` sections. It still rides along as detail on a revision that earned its line some other way.
- **Creations collapse to a daily count.** 1456 rows naming every fact ever added buried the revisions a reader came for. The growth signal is the part a changelog must not lose and is fully preserved — 1456 facts still reported added, now in 59 rows. *Which* facts arrived is what the directory indexes and each document's `generated.at` are for.
- **Correction:** the Update counts in this table were inflated by a defect found in the sixth pass — 179 of the 260 were rows wrongly labelled `— created`. The true count is 81.
- **Update rows carry their delta.** They rendered as a bare `**Update** <title>`, asserting that something happened while unable to say what. 62 of 322 had no reportable delta at all (a retag, an origin change) and are dropped; the other 260 gained wording.
- **The month index is a jump bar, not headings.** §9 requires a *flat* list of date-grouped entries with ISO `YYYY-MM-DD` headings, so grouping by month the way `views/` does would read better and break both rules. Pinned by a test that fails if `##`-level month headings or `###` day headings appear.
- **An empty log produces no file.** `# Log / No changes recorded.` was the one place in this package writing an empty document; `renderRetired` and the digests already declined to.

Superseded by the sixth pass below: `log.md` records creations of facts since retired — 1456 Creation events against 1208 live facts. Defensible for a changelog, and `views/retired.md` covers the withdrawals, but the log names documents that are not in the bundle.

### Sixth review pass — the changelog is not one file

§9 allows a `log.md` **at any level**, recording changes to that scope, and names `**Deprecation**` as one of its three conventional labels. The bundle used neither: one root log carried everything, and **247 withdrawn facts appeared in `views/retired.md` and nowhere in the changelog**. A knowledge base reporting only what it gained was describing half of what happened to it.

Every event now lands in **exactly one** log. A revision or a withdrawal belongs to the folder holding the fact it happened to; when that folder is gone — its last fact retired — it falls to the root, the only scope that still contains it.

|  | folder logs | root log | before |
|---|---:|---:|---:|
| Creation | 0 | 59 rows (1456 facts) | 1456 rows |
| Update | 68 | 13 | 322 |
| Deprecation | **153** | **94** | **0** |
| files | 88 | 1 | 1 |
| bytes | 42015 | 18241 | 176850 |

- **Creations stay at the root**, and are the one exception to the partition. They are 93% of all events, so splitting them would empty the root of everything but the deleted corner of the base: it would have kept 215 rows describing only facts whose folder no longer exists, while the 1362 actually added scattered into 922 folder files — 747 of them saying nothing but `1 fact added`, which the folder's own `index.md` and each document's `generated.at` already say.
- **A successor is named, never linked.** The same Deprecation row renders into the root log and into a folder log at a different depth, so a relative link correct in one is broken in the other. `views/retired.md` remains the view that links.
- **An `**Update**` row could claim `— created`.** Revision 0 is the creation slot, labelled `created` because nothing precedes it to diff against — but when the walk never reaches a path's real creation, that path has no `Creation`-marked event, the normalization turns every one of its events into an `Update`, and the label landed on one. It asserted a birth on exactly the facts whose real birth is already lost. Reproduced by lowering the walk bound, and confirmed red against the code it replaces.
- **That defect inflated the fifth pass's numbers.** 179 of the 260 "meaningful updates" were those rows; the true count is 81 — where the first estimate had it, before the bug moved it.

### Seventh review pass — a race with git's own maintenance

Found by replaying this PR's own documented workflow — commit a publisher
`README.md`, then sync — against a live base. It failed **8 runs in 12** with a
bare `object not found`, on a repository `git fsck` calls intact. With
`gc.auto=0`, 0 in 3.

- **A background repack made objects unreadable mid-run.** git runs `git maintenance run --auto` after ordinary commands like commit, detached, and it writes loose objects into a new packfile and then deletes them. go-git builds its packfile index ONCE, on first use (`storage/filesystem.ObjectStorage.requireIndex`), and never rescans — so an object that moved into a pack created after that point is gone from the loose directory and absent from the cached index. Reachable on the happy path, not a corner: each changed sync writes ~1000 loose objects through go-git, which never packs them, so the repo drifts toward git's threshold and the user's own `git commit` is what fires the repack. The fix rescans (go-git documents `Reindex` for exactly this) and retries once — provably enough, because git writes the new pack before it prunes the loose copy, so every object is readable in one place or the other at every instant. A second failure is therefore not a repack, and is reported rather than looped on.
- **The same race has a second face.** `Tree.Tree` converts `ErrObjectNotFound` from a subtree read into `ErrDirectoryNotFound`, discarding the cause, so a repacked tree arrives under a name that says nothing about objects. A recovery keyed only on `ErrObjectNotFound` walked straight past it — caught by a live run that failed with `directory not found` after the first version of the fix was already in.
- **Recovering in the wrong unit cost more than the failure.** Wrapping the whole export worked but took every affected run from **2.9s to 5.6s**, redoing a full history walk to recover a failure whose last step was all that needed repeating. The commit tail — the likeliest loser, since it runs last and right after the user's git command — is now guarded on its own, with the whole-export retry left as the net for the stages that read thousands of objects. Back to 3.0s, 8 of 8.
- **A recovery that ended in a sentinel went unannounced.** Recovery means "the object became readable", not "the operation succeeded": a retried commit legitimately ends in `ErrEmptyCommit`, which is how this tool says nothing changed and is the most common re-sync outcome. Keying the note on a nil error made every recovery on that path silent, so a run repeated work and never accounted for it — the same failure this branch keeps indicting, found in the fix for it.

Reviewing the fix then found three more, in the fix itself:

- **A recovery announced from inside an open stage deleted that stage's result.** `ui.Note` closes whatever stage is open, and a closed stage's `Done` is a no-op — so `branches` recovering mid-comparison printed the note and swallowed the `✓ Comparing N branches` line it exists to produce, while claiming *"the stages above ran twice"* of a stage that had not printed at all. Both sites now note after the stage closes, and the trap is pinned by a UI test rather than left as something the next caller has to know.
- **`errors.Is` could not see the failure on the most-executed path.** go-git's tree diff wraps with `fmt.Errorf("from: %s", err)` — `%s`, not `%w` (`utils/merkletrie/doubleiter.go`) — destroying the chain outright, and every `wt.Status` goes through it. So `sync`'s own up-to-date check produced `from: directory not found` that no amount of sentinel matching could catch: 1 failure in 6 runs with the typed matching already in. Classification now prefers the typed errors and falls back to a narrow substring — the same shape the host-key classifier already uses, for the same reason, and the cost of the fallback being wrong is one wasted retry rather than a wrong answer.
- **`ownedPathsClean` was unguarded** — the `Status` call above, on the path a scheduled `sync` takes every time it has nothing to do, which makes it the most-executed object read in the tool.

Ten tests across `tools/okf/repack_test.go` and `ui_test.go`, including a real reproduction that packs objects through a second repository handle and deletes the loose copies. Each confirmed red against the code it replaces, including **both directions** of the substring fallback: dropping it loses the chain-breaking wrap, widening it swallows unrelated failures. The reproduction asserts the stale-cache miss happens *before* exercising the recovery, so it fails loudly rather than passing vacuously if go-git changes upstream.

Verified **10 of 10** on the live workflow, against 8 failures in 12 before any of this.

## Test coverage

- **`internal/okf`** — goldens (including a second *feature-rich* fixture covering `# History`, `# Cited by`, `views/retired.md`, digests, entity hubs, `sources:`, `status: draft`, bracket escaping), determinism, conformance, round-trip.
- **`internal/okfsource`** — in-memory fixture repos: fact reading, the history walk, revision ordering, retirement classification, both merge cases, ontology loading and its embedded-default fallback.
- **`tools/okf`** — against a local fixture KB: clone produces a valid bundle on a correctly-named branch; an unchanged `sync` makes **no commit**; a retired fact's document is **removed**; a publisher's `README.md` survives; source refs are absent from a pushed remote; `--publish-source` controls the `source:` field; a second branch gets its own orphan; `branches` lists exported and unexported alike, counts drift, describes branches without checking them out, and flags an export whose source is gone.
- **UI** — piped output stays plain and line-based; a stage is visible before it completes.
- **Degradation reporting** — a walk cut at a lowered bound warns and names the damage, while one inside the bound stays **silent** (a warning that is always on is a warning nobody reads); an unparseable fact is reported and a `README.md` under `kb/` is not; `looksLikeFact` is table-driven over both signals and their near-misses.
- **Recovery paths** — a publisher's edit blocks a branch switch *and is named*, an untracked file does not block, a dirty bundle file is repaired rather than refused, and `restoreOwnedFromIndex` refuses a path outside the owned boundary; a failed clone removes a directory it created, empties one it was given, and **retries successfully**; each transport/flag mismatch errors while the matching flag and ambient environment still work.
- **Regression** — `TestVerify_NoGeneratedRefs` guards the Critical that motivated all of this; `TestFetchSource_DoesNotRewriteTheStoredRemote` and `TestCountNewCommits_DoesNotCrossIntoAncestorsOfStop` pin the two review findings above; `TestCLI_IncrementalStagingMatchesFullRestage` requires an incrementally staged sync to produce the same tree as a full re-stage, and `TestCLI_UnchangedSyncStagesNothing` pins the no-op case. All fail against the behaviour they replaced. Every test added in the second review pass was checked the same way — nine mutations, each reverting one fix, each caught. The third pass adds `TestCLI_OrphanBranchInheritsThePublishedSource` (and its privacy converse), `TestCLI_ANewReleaseRerendersAnUnchangedSource` (which pins both directions: a new release re-renders, a rebuild of the same release does not), `TestBranches_ReportsTwoOutputBranchesClaimingOneSource` and `TestBranches_ColumnsAlignAcrossMultiByteNames` — all four confirmed failing against the code they replaced. The fourth pass adds `TestCLI_FailedOrphanSyncRestoresTheRepository`, `TestCheckoutOutputBranch_UnbornHEADStillClearsTheIndex`, `TestReconcile_RefusesASymlinkedOwnedRoot`, `TestCLI_SyncAndBranchesWorkFromASubdirectory`, `TestCLI_SyncRefusesARepositoryItDidNotCreate`, `TestExplainFetchError_HostKeyMismatchNeverSuggestsKeyscan` (with its unknown-host and unrelated-handshake converses), `TestCheckAuthApplies_UsernameWithoutATokenIsRejected`, `TestBuild_HubKeyNamedIndexDoesNotClaimTheDirectoryIndex`, `TestBuild_RepeatedRefIsPresentedOnceButRoundTripsVerbatim`, `TestRoundTrip_AuthoredHeadingNamedLikeAGeneratedOne` and `TestStripGenerated_PeelsOnlyTheTrailingSections` — each confirmed red against the behaviour it replaces, and each of the four commits verified green in isolation.

## Known follow-ups

- `src://` → forge permalinks once a slug→repo mapping exists.
- Per-fact `description` — needs an authored summary field on the fact format.
- Incremental rendering is out of scope: rendering is now a small fraction of a run (41 ms of an 18 s clone on 2816 documents — the cost is the commit walk and git staging, not the mapper), while incremental view maintenance would add substantial complexity, since a title change ripples to every citing document.
- New output branches are created as **orphans**, so a publisher's `README`/`LICENSE` do not carry across from an existing branch. Deliberate — the bundles are unrelated snapshots — but worth revisiting if it annoys in practice.

https://claude.ai/code/session_01HcWDvAy9V8e1yG7bGU1ifv


knomit:
### Code review

PR #49, opened 2026-07-29
feat(backup): litestream replication so a cold boot is fast

knomit:
## knomit-backup — continuous replication so a cold boot is fast

Replicates knomit's SQLite state to object storage continuously, and rehydrates it on startup, so a replaced container or a fresh volume comes up with its data already there instead of re-cloning every repo from origin.

```toml
[backup]
enabled  = true
url      = "s3://my-bucket/knomit"
instance = "prod-1"
```

That is the whole configuration. `knomit serve` then restores `control.db` and every registered repo database before it opens anything, and replicates them from that moment on.

---

## The replica is a CACHE, not durable state

This is the design constraint everything else follows from, and it is worth stating plainly because the branch originally got it backwards.

`kb/principles/philosophy/git-is-the-only-source-of-truth` says the git repository is canonical and every SQLite database knomit keeps is rebuildable from it. So the replica is a **warm-start cache**: losing it costs boot *time*, never data. A repo that cannot be restored is re-cloned from the origin its registry row records.

**Consequence: nothing here can refuse a boot.** `backup.Open` failing, a restore failing, an unreachable bucket — all log at ERROR and the server starts anyway. A cache miss must never become an outage.

The first draft of this branch was built the other way round, as though the replica were the durable copy, and most of its complexity was machinery defending it. All of it is gone:

| Removed | What it did |
|---|---|
| `ErrRestoreIncomplete` | refused the boot when any repo failed to restore |
| `StrictMissing` / `ErrRepoUnrecoverable` | refused the boot when a repo had no database and no origin |
| `Preflight` / `ErrDiverged` | refused the boot when the replica was ahead of local |
| `internal/homelock` | refused a second server on one `KNOMIT_HOME` |
| `knomit restore` | operator command to overwrite a live database |
| `reset --force` | gated `reset` behind a flag when backup was on |

`Bootstrap` returns no error for any replica problem, and `restoreHome` returns nothing at all, so a refusal cannot be reintroduced without changing a signature. Rationale is recorded in `kb/decisions/backup/warm-start-cache-not-durability`, including the one boundary it does not cross — a repo created *without* an origin keeps its only copy of its git history inside that `.db`.

**This PR adds no new CLI command.** `root.go` registers the same five it always did.

---

## Why replication runs in a separate process

litestream v0.5 drives SQLite through `modernc.org/sqlite`; knomit drives the same files through the cgo `mattn/go-sqlite3` build and cannot switch, because sqlite-vec has no modernc build.

Two SQLite *builds* in one process do not see each other's file locks — POSIX advisory record locks do not conflict between descriptors held by the same process, and SQLite's compensating per-process inode table belongs to a single build. Demonstrated, not theorised: with litestream holding a read lock, knomit's close deleted its `-wal` and unmapped its `-shm`, 3 runs out of 3. One fact explained three symptoms that had been chased separately.

Across processes those locks work exactly as SQLite intends. So `knomit-backup` is a child process; knomit speaks newline-delimited JSON to its stdin/stdout and reads its stderr into the server log. It exits on EOF, which is also what happens if knomit is SIGKILLed — no orphan agent can outlive its parent.

```
internal/backup/            knomit-side client: Manager, process supervision
internal/backup/proto/      the wire protocol, spoken by both sides
internal/backup/agenttest/  builds the agent binary for tests
tools/backup/               the knomit-backup binary
tools/backup/agent/         the litestream half
```

The agent lives under `tools/` deliberately: nothing reachable from the knomit binary can import it, so "knomit never links litestream" holds structurally. `go list -deps .` returns zero litestream/modernc packages; `TestKnomitBinaryDoesNotLinkLitestream` is now a backstop rather than the only guard.

---

## Boot sequence

1. **identity** — resolve `agent_name` + SSH key. First, because an unstable identity would restore one branch's history and then write to another — a silent fork that no amount of re-cloning fixes.
2. **backup** — build the replica client. Failure logs and continues.
3. **control.db** — restore. Must precede step 4: opening the registry *creates* `control.db` when absent, which would leave restore nothing to fill.
4. **registry** — read the intended repo set. A restored machine's `repos/` is empty, so the registry is the only record of what should exist.
5. **repos** — restore every intended database that is absent locally.

Then `app.New` runs `repos.Manager.Start`, and only what actually opened is registered for replication — the registry says what *should* exist, but only `Start` knows what did.

### The repo registry

`Manager.Start` used to glob `repos/*.db`, which returns nothing on a restored volume — so the server silently came up with no repos. A new `repos` table in `control.db` is now authoritative, holding each repo's name plus its origin URL/branch. Each row is reconciled against disk: present → open, absent with a recorded origin → re-clone, absent with nothing to rebuild from → logged and skipped.

This is also what makes "point it at the proper origin" work at all, and it is why the registry survived the simplification while nearly everything else around it did not.

Two related changes ride along: **the default `core` repo is gone** (zero repos is a valid state; knomit creates nothing implicitly), and `verify`/`reset` therefore require an explicit `--repo`/`--name`.

---

## What is replicated

- `control.db` — the registry itself
- every open repo database
- **archived** repo databases, under an `archive/<ksuid>` prefix with retention **disabled**

The archive case is a deliberate exception to the cache framing — archived repos are never served and never participate in a boot, so replicating them is durability, full stop. Kept on purpose; see `kb/decisions/backup/keep-archive-replication`. Note that prefix accumulates objects forever by design, and `Purge` is the only thing that reclaims them.

## Config

`enabled`, `url`, `agent_path`, `instance`, `snapshot_interval`, `snapshot_retention`, `l0_retention`, `monitor_interval`, `restore_timeout`, `status_cache_ttl` — with `KNOMIT_BACKUP_*` env overlays for the first four plus the two timeouts.

There is deliberately **no `auto_recover`**: litestream resolves LTX errors by resetting the replica, which discards history to paper over a condition worth seeing.

Backup is compiled **out** of the desktop build (`internal/config/backup_desktop.go`) — not defaulted off, compiled out, so no config file or environment variable can turn it on there.

## Testing

26 new source files, 36 new test files. `go test ./...` and `go test -tags desktop ./...` both green; CI's desktop job now runs the tagged suite over the whole tree rather than just `./tools/desktop/...`, which is what actually executes the desktop backup-off guard.

Story tests cover the end-to-end wipe-and-recover path against a real agent binary built from the working tree, and the production shutdown ordering (knomit's SQLite handles close first, then the agent's final sync).

## Fixed in review

Two silent-loss shapes found reviewing the branch, both in `internal/repos`.

**A swapped-out database's companion files were deleted, not set aside.** This was previously listed here as a known issue and deferred; the review found it reaches one path further than described, so it is folded in.

They have to leave `ri.dbPath` before the staged replacement is renamed in — a `-wal` surviving onto the new file is replayed into it by SQLite on first open, and a WAL header carries no database identity, so the corruption is silent. Deleting them does that, and is safe only if the pre-swap database is never wanted again. It is wanted on two paths: the rename can fail, and the reopen after it can fail into `rollback()`. In exactly the case the step exists to cover — a close that did not checkpoint — a delete has already taken the original's un-checkpointed tail with it, so `rollback()` restores a `.bak` missing those writes *and* a failed rename reopens the original without its tail even though nothing was replaced. Neither raises anything: an incomplete database opens fine.

The companions are now **moved** beside the `.bak`, so it is a complete database, and both recovery paths move them back. The destination is cleared whether or not there is anything to move — the common case is a clean close with nothing there, which is precisely when a companion left by an earlier swap would survive to be paired with the database arriving next.

**An archive rollback could leave a repo live but unregistered.** `Archive` captures the active registry row so a failure can put it back, and restored it only `if hadPrior`. When the capture found nothing, or failed, nothing restored a row at all — `reinstateLive` re-opens the repo and re-tracks it for replication, and neither writes to the registry. Since `Start` reads the registry and no longer globs the disk, the repo served normally until the next restart and then simply was not there. The `!hadPrior` arm now calls `EnsureActive`, and `reinstateLive` calls `RecordOrigin` so the origin is re-derived from the store.

Both fixes carry regression tests, each confirmed failing without its fix. The archive one injects the failure in SQL — a trigger that rejects inserts of archived rows only, so recording the archive fails while the rollback's own insert still works.


knomit:
Will not merge, stale implementation

PR #58, opened 2026-08-03
UI: readable refs, a location header for the Library, and back navigation

knomit:
Two related bodies of UI work: making fact references readable, and moving the
Library's location into the Library.

## References

`src://7b4887ce51d9/web/src/EdgesRail.tsx@<40-hex>:<40-hex> (source)` was ~110
monospace columns, nearly all hex. It now reads

    src://7b4887ce51d9/web/src/EdgesRail.tsx@b27972ca…:f08b5ecf…

with the full citation on hover and a copy button that puts a runnable
`git cat-file blob <the actual hash>` on the clipboard — the tooltip used to say
`<blob>` and leave you to retype 40 characters. A ref with a commit but no blob
offers `git show <commit>:<path>` instead, and says it fails if the path did not
exist at that commit; a ref with neither offers nothing rather than something
unrunnable.

The repo id is replaced by a name only when it names a MOUNTED repo, which in
practice means `kb://` refs. A `src://` id is the source repo's root commit — a
different namespace from KB-store ids, with no registry anywhere on either side
of the wire — so it is left as written.

This shortening was first built server-side and reverted (e95ca915). The
invariant it was justified by governs ref CLASSIFICATION, not formatting; and
the server's name lookup could never have fired, because it searched the KB-store
id namespace. That's now recorded as a gotcha rather than rediscovered.

## Console panel

Deleted, with its store, ring buffer and CONSOLE_* actions. Nothing addressed to
a user was ever routed there. The footer stays as StatusFooter (position, task,
version) and is no longer a control. Lines that were addressed to a person moved
to SET_NOTICE; the rest go to the browser console via `diag`, which exists so
that an SSE stream that silently stops — the one failure this app cannot
otherwise show — stays diagnosable.

## Connections rail

Hidden for a fact with no edges either way. The rail owns the fetch and App owns
the column width, so the rail reports and App decides, keyed by the anchor the
report describes. Not reported on error: an unreachable server is not the same
answer as "this fact stands alone".

## Library navigation

The location moved out of the `path:` filter chip — which renders in the RIGHT
column, across the splitter from the list it scopes — and into the Library's own
header:

    [<]  kb / decisions / ui        2  [tree][clock]
         ref-display

Long chains collapse behind a `…` menu (extracted from TrailBreadcrumb and
shared, not copied), dropping the root before the parent when narrow. Entering a
directory is now one action instead of two. Back is `⌘[` / `Alt+←` as well as
the chevron, and the Library remembers your scroll position per folder.

Back only, deliberately: `selectTrail` and `jumpTrail` read the same `navStack`,
so a non-destructive cursor for Forward would grow phantom time-travel crumbs.
That coupling is documented in the KB.

## Notes for review

- No Go changes. `internal/` is byte-identical to `dev`.
- 653 web tests pass; lint is at the pre-existing 89 (down from 97 on `dev`).
- Two bugs found in the browser that tests could not see: `Alt+←` moved two
  levels because App and Library both listen on `window`, and the header
  changed height between root and scoped by 2.25px.
- `FilterBar — path chip breadcrumb`'s two tests were MOVED to
  `LibraryHeader.test.tsx`, not deleted.


knomit:
### Code review

Found 2 issues:

1. A failed edges fetch is unreachable in the UI, so a server error looks identical to "this fact stands alone". On failure `useFactEdges` clears both edge arrays *and* sets `error`:

https://github.com/knomit/knomit/blob/b80737858cf75d48a4f801dc6258a96d7b4fd099/web/src/useFactEdges.ts#L76-L81

`ConnectionsCell` then gates interactivity purely on the count, so both cells render as inert `<div>`s with no `onClick`:

https://github.com/knomit/knomit/blob/b80737858cf75d48a4f801dc6258a96d7b4fd099/web/src/ConnectionsMenu.tsx#L26-L28

But the error text only renders inside `ConnectionsPanel`, which is only openable by clicking an interactive cell — and `edgesError` is never forwarded to the cell:

https://github.com/knomit/knomit/blob/b80737858cf75d48a4f801dc6258a96d7b4fd099/web/src/ConnectionsPanel.tsx#L137-L139

This contradicts the hook's own contract (`/** Fetch error, or null. Empty edges are NOT an error. */`, [useFactEdges.ts#L12](https://github.com/knomit/knomit/blob/b80737858cf75d48a4f801dc6258a96d7b4fd099/web/src/useFactEdges.ts#L11-L13)) and the PR description's stated goal that a rail is "not reported on error". No test covers it — `ConnectionsPanel.test.tsx` renders the panel with `open` forced, bypassing the gate.

2. A comment added by this PR describes `edgeAnchorCommit` in terms of `EdgesRail`, which the same PR deletes (`web/src/EdgesRail.tsx` and both its test files are removed):

https://github.com/knomit/knomit/blob/b80737858cf75d48a4f801dc6258a96d7b4fd099/web/src/state.ts#L488-L490

PR #61, opened 2026-08-04
Repo registry: control.db is authoritative for what repos exist

knomit:
## The control-plane registry from #49, without litestream

Extracts the repo registry — and the default-repo removal it forces — out of #49, so that PR can be closed without losing this. **Nothing here replicates anything.** No `internal/backup`, no `tools/backup`, no `app.Bootstrap`, no `BackupTracker`, no config keys. `app.New`'s signature is untouched, so the desktop build is untouched.

---

## The problem

**The filesystem was the registry.** `Manager.Start` globbed `repos/*.db` and opened whatever it found. Three consequences, none of which need a replica to matter:

1. **A repo's origin lived only inside its own database.** Lose the file and nothing anywhere records where it came from. The one thing that could rebuild it is the one thing that went missing with it.
2. **Archive metadata lived in `repos/archive/<id>.json`** — a sidecar next to the SQLite file, with its own parse-failure and partial-write modes, describing control-plane state that had no business being outside the control plane.
3. **One repo was privileged.** `core` was created implicitly at boot by `initDefaultGit`, outside every registration path, driven by a *server-wide* `git.origin`. That produced asymmetries that had nothing to do with defaults: a cloned repo on a server without `git.origin` never index-synced its upstream and never reconciled at boot, because both were gated on the server-wide key rather than the repo's own remote.

A new `repos` table in `control.db` answers "what should exist, and where did each come from?" — the question an empty disk cannot answer. Every repo then arrives through one path, which is what makes recovery uniform.

## `control.db` gets versioned first

It had **three tenants creating their own schema at open time**: a `lensSchema` const, a `repoSettingsSchema` const, and a runtime `ALTER TABLE lenses ADD COLUMN description` that swallowed `duplicate column name`. Adding a fourth tenant to that was not on.

`migrate.Control` is a separate migration set from `All`/`Core` — different file, its own `schema_migrations`, stock `sqlite3` driver since `control.db` holds no vectors. Migration 1 is the pre-`description` schema with `IF NOT EXISTS`, so an install created before versioning is **adopted** rather than failed. Migration 2 is that `ALTER`; where the column already exists, the existing `upWithRecovery`/`alreadyApplied` path marks it applied — the versioned equivalent of the error the runtime code used to swallow. Verified against a real pre-migration `control.db`; see **Rehearsals** below.

## The registry

`repos(name, origin_url, origin_branch, state, archive_id, created_at, archived_at)`, primary key **`(name, archive_id)`**.

The composite key is not tidiness. *A repo name is unique only among ACTIVE repos.* Archiving `work` and creating a new `work` is supported and tested, as is archiving the same name twice — so with `name` alone, the new active row overwrote the archived one and the archived repo became unrecoverable the moment its name was reused. Active rows carry `archive_id = ''`, so a name still has at most one active row and upserting an active repo by name still updates in place.

Three API decisions, each with a defect behind it:

- **`EnsureActive` is not a convenience wrapper over `Upsert`.** `Upsert` is a whole-row write — every column comes from `excluded` — so upserting a partially-filled record **blanks the origin and restamps `created_at`**. Right for a repo being created, wrong for one being *re*-registered. `Rescan` adopts a database that may already have a row (`Start` skips-and-logs a repo whose open failed, leaving its row behind), and blanking `origin_url` there destroys the one field the rebuild path reads. `EnsureActive`'s conflict arm touches `state` and `archived_at` only.
- **There is deliberately no `SetState(name, state)` and no `Delete(name)`.** Both read as if a name identified one repo, which stopped being true at the composite key. `SetState("work", archived)` would flip every row sharing the name and leave the active one with an empty `archive_id` — an archived row no `Restore` or `Purge` could ever address. Archiving is "insert the archived row, retire the active one": `DeleteActive` + `DeleteArchive`. The comment saying so is load-bearing.
- **`DeleteArchive` is keyed by id, never by name** — the archived repo's name may since have been claimed by a live active repo.

## Startup

`adoptFromFilesystem` runs **before anything opens**, and only when the registry is empty: on an upgrade the disk is the only record of what exists, so the filesystem gets one last say. Afterwards the registry is authoritative and the disk is never consulted again. It folds `repos/*.db` into active rows and `repos/archive/*.json` manifests into archived ones.

`Start` then reconciles each active row against the disk: **present** → open; **absent with a recorded origin** → re-clone through the ordinary create path (so the clone inherits `setupIndex`'s both-branch sync rather than reimplementing it); **absent with nothing to rebuild from** → logged and skipped.

**No single repo can fail the boot.** Every per-row outcome is logged at ERROR and skipped — an open that failed, a clone that failed, nothing to rebuild from. One repo's unreachable origin or corrupt database must not take the other repos on the instance offline with it. The row survives in every case, so the next restart retries and the state stays diagnosable. `Start` returns an error only for whole-manager conditions: the repos directory, `control.db`, the session reaper's config. The ERROR for a failed re-clone spells out the `sqlite3 … DELETE FROM repos` recovery, and notes that a token-authenticated origin can never be re-cloned this way — the token was encrypted *inside* the database that went missing.

`isValidRepoName` is now applied on the **read** paths too (the reconcile, `Rescan`, adoption), so a hand-edited or corrupted row skips-and-logs instead of refusing the boot.

## Origin write-through

`RecordOrigin` copies a repo's live origin from its store into its registry row. The store's `remotes` table stays the source of truth; the row is a copy for the one case the store cannot serve — the store file is gone. A copy is only worth having if it is kept current, so **every path that changes an origin calls it**: clone-mode `Create`, the boot reconcile, `Rescan`, `PUT`/`PATCH`/`DELETE` on `/repos/{repo}/origin`, and both origin-session commit paths that write the remote directly rather than through the provider. Without the last of those, an origin attached *after* creation lived only in the store — the exact hole the registry exists to close.

It is **authoritative**, which means it CLEARS the row when the store reports no origin. A disconnect is an origin change like any other, and a registry that kept the old URL would have the next boot re-clone from a remote the user deliberately removed.

The subtlety is `originOf`'s third return, and conflating the two empties is a bug in both directions:

- `ok=true, url=""` — the repo **genuinely** has no origin. Bring the registry into line, *including clearing* what it holds.
- `ok=false` — the store could not be READ. Nothing was learned; leave the row exactly as it is. Treat this as "no origin" and a transient read failure erases the one field a rebuild needs.

The `PUT /origin` write-through is placed **before** `ActivateSync`, deliberately: that call can fail the request with a 502 while the origin *is* already persisted, so a write-through after it would be skipped for exactly the origin most likely to need re-cloning later.

## Archive / restore / purge

The registry row **replaces** `repos/archive/<id>.json`. `Archive` inserts the archived row and *then* retires the active one — the window where both exist is recoverable, the window where neither does is not. `Restore` mirrors it: active row first, then `DeleteArchive`, carrying `CreatedAt` across so a restored repo does not look newly created. `Purge` deletes by archive id. `ListArchived` reads rows; the `ArchiveInfo` shape and newest-first ordering callers see are unchanged. A leftover legacy manifest is removed on restore/purge — left in place, a later adoption on a machine that lost `control.db` would resurrect a dead entry pointing at a database that is gone.

**Fixed here, carried over from #49:** an archive rollback could leave a repo **live but unregistered**. The rollback restored the captured active row only `if hadPrior`; when the capture found nothing or failed, nothing restored a row at all. Since `Start` reads the registry and no longer globs, the repo served normally until the next restart and then simply was not there, with nothing in the log connecting the two — and `RecordOrigin` could not repair it either, since it declines to write when the row is absent. The `!hadPrior` arm now calls `EnsureActive`, and the three recovery sites collapse into `reinstateLive`, which re-derives the origin from the store now the repo is open again. The regression test injects the fault in SQL — a trigger rejecting inserts of *archived* rows only, so recording the archive fails while the rollback's own insert still works. Confirmed failing without the fix.

## The default repo is gone

This is forced, not opportunistic: `core` was the last path by which a repo could exist without a registry row.

**Deleted:** `config.DefaultRepoName`; `GitConfig.Origin` and its `KNOMIT_GIT_ORIGIN` overlay (only ever applied to `core`, gated on `isDefault`); `repoBuilder.isDefault` and every branch on it; `initDefaultGit`; `resolveOriginUpstream`; `ErrCannotArchiveDefault`; `ErrCannotArchiveLast`.

**Three things moved rather than vanished**, because deleting them outright would have silently dropped behaviour:

- **Remote-HEAD detection.** It lived in `initDefaultGit`; the clone path recorded a hardcoded `"main"` when no branch was requested, so a master-default origin got refspecs pointing at a branch that does not exist. `detectUpstream` now runs in `initClone` *before* the clone, so clone, refspecs and the `remotes` row all name one branch.
- **The ontology seed.** `initClone` passed no seed files. That only matters for an *empty* origin — a non-empty one brings its own — but a repo cloned from an empty remote had no `domains/ontology.yaml` at all.
- **`upstreamMain` and the startup reconcile.** Both were gated on the server-wide origin. `resolveUpstreamMain` reads the repo's own stored remote; empty means no origin. `warnIfUpstreamMissing` is a loud diagnostic and deliberately not a self-heal — re-detecting every boot costs a network round-trip per repo and could overrule a deliberate choice.

The prefer-`main` rule also moves out of `InitFromRemote`'s guard into `detectFromRemote`. Once `initClone` resolves the branch itself and passes it in **non-empty**, that guard is skipped entirely — so a remote whose symbolic HEAD points at an agent branch (its GitHub default set to `agent/<host>`) would otherwise make that agent branch the consensus upstream.

**Archiving the last repo now succeeds** and leaves zero repos. A user who created a repo by mistake has to be able to undo it, and `Restore` brings it back.

**`knomit reset` is removed.** It deleted a repo's `.db` directly, which predated the registry; once the registry became authoritative the row survived the delete and the next boot either re-cloned the repo straight back — reset as a no-op with extra steps — or logged it forever as registered-but-unrebuildable. Removal goes through the registry now: `DELETE /api/v1/repos/{repo}` to archive, `DELETE /api/v1/archived/{id}` to purge. `app.ResetRepo` had no other caller and goes with it. `root_test.go` pins the command set **exactly**, so a command added without a decision behind it is visible in the same place.

`knomit verify` and `knomit-bridge` now require `--repo` (or `--all`), validated *before* `app.New` so a missing flag is not buried behind seconds of boot. `verify --all` with no repos says so rather than exiting 0 in silence, which read as "everything verified clean".

## Zero repos is a first-class state

A fresh install has none until the user makes one, so the empty screen must be a way **in**: it renders the repo manager alongside it and creates the first repo right there, instead of telling the user to go run a command line and reload. `RepoManager` lands on the create form when there is no current repo — via a `homeView` rather than `{kind:'repo', name:''}`, which is a non-null selection that slips past the `sel ??` fallback and renders a detail pane fetching `/api/v1/repos//origin`.

## What was dropped from #49 on the way

- `Deps.StrictMissing` / `ErrRepoUnrecoverable`. It made `Start` refuse the boot for a repo with no database and no origin, and existed purely for the replica: with replication running, omitting such a repo meant the empty local state got replicated over the good backup. Without a replica that reasoning does not survive, and what is left is a flag that turns one repo's bad day into a whole-instance outage. The rule is now unconditional skip-and-log.
- `Manager.OpenDBPaths` / `ArchivedDBPaths` — they existed only to feed the tracker.
- Reserving the repo name `control`. Its three stated consequences are all replica-namespace consequences; `<home>/control.db` and `<home>/repos/control.db` do not collide on disk. Worth doing on hygiene grounds, but as its own change with honest reasoning, not smuggled in here.
- The swapstore staging / WAL-set-aside durability fixes. Real and independent of both backup and the registry; they belong in their own PR.
- The `reserveNameAndOrigin` wrap around `Archive`. It closes a genuine Create-during-Archive race on the `.db` file, but #49 justifies it entirely in tracker terms. Left out rather than ported with reasoning that no longer applies — noted here so it is not lost.

Comments that explained the registry via litestream are reworded to the reason that stands on its own.

## Rehearsals

Beyond the suites, run by hand against a **real pre-migration `KNOMIT_HOME`** — a `control.db` carrying `lenses` (with `description`), `lens_reads` and `repo_settings`, and no `schema_migrations`:

- **Upgrade.** Migrations 1–4 apply; migration 2 hits its own committed body and completes through `upWithRecovery` (`migration collided with its own committed body; completing an interrupted recovery`); the repo is adopted and verifies CLEAN; a stray `core 1.db` is skipped for an invalid name; a legacy archive manifest becomes an archived row with its origin and `archivedAt` intact; 3 lenses and the settings table survive untouched.
- **Idempotence.** Second boot logs no adoption at all and leaves every row unchanged.
- **Boot resilience.** A row pointing at an unreachable origin with no database, plus a row with neither — the server starts, serves the healthy repo, logs both at ERROR with the recovery spelled out, and leaves both rows in place.

## Verification

`go test ./...` and `go test -tags desktop ./...` green. `gofmt` clean on every file touched (four pre-existing `gofmt`-dirty files on `dev` left alone). `tsc --noEmit` clean; 702 web tests across 61 files pass.

## Follow-up

Two KB facts to write once this lands: that `control.db`'s `repos` table is authoritative and nothing may reintroduce a `repos/*.db` glob as a source of truth (`adoptFromFilesystem` being the sole, registry-empty exception), and that every origin-changing path must call `RecordOrigin`.


knomit:
Closing in favour of #65, which now carries this work.

#65 was stacked on this branch, so it already contained all 17 commits here, unchanged and in order. Rather than keep a two-PR stack for one migration, #65 has been retargeted onto `dev` and merged with it, and now presents the whole control.db migration as a single PR — the registry from this branch as Part 1, the origin credentials as Part 2.

Nothing was squashed or rewritten: every commit from this branch is in #65 with its original hash (`aa22f3f5` … `11c4e44a`). The review discussion on this PR stays here and stays readable.

The `feat/repo-registry` branch is left in place; it is an ancestor of `feat/origin-credentials-in-control-db`, so it can be deleted whenever convenient.

PR #62, opened 2026-08-04
Overview highlights: rank facts by what others were built on

knomit:
The overview answered "how much is in here?" with five count tiles. Nobody
arrives at a knowledge base with that question. This replaces the bottom of that
page with **Highlights** — the facts the most others were built on — and
compacts the tiles into one strip to make room.

Impact is *earned* rather than self-reported: confidence is a number the
authoring model assigned itself, while `DERIVED_FROM` out-degree is evidence of
work done. On `core` the distilled layer out-derives observations 27.6x, so
excluding observations falls out of the ranking instead of a hand-maintained
blocklist.

The two click targets are the whole interaction: a type pill filters, a title
opens the fact.

## What's here

- Compact stat strip; domain/entity chips unchanged — they were already the only
  working controls on the page.
- Top-10 highlights with an `Impact · Confidence · Recent` control. Ranking is
  server-side over the full eligible set, so switching axes refetches. The client
  never re-sorts: the top-10 under one axis is not the top-10 under another, and
  sorting a truncated list would contradict the caption.
- The default axis is chosen per repo by separation ratio (top layer vs
  observations, threshold 3x). Impact ranking is useless on authored corpora —
  `knomit-kb`'s observations out-derive its policies — so those fall back to
  confidence.
- Same treatment on the lens union, deduped write-first via
  `federate.WriteFirstWinners`.
- `/stats` and `/lenses/{lens}/stats` gain `types`, `highlights`, `default_axis`
  and an `axis` parameter, with OpenAPI to match. The repo `path` parameter was
  undocumented despite scoping every existing aggregate; documented now.

## Two defects the whole-branch review caught

Both had passed per-task review.

**Impact was inflated for 81% of facts.** It counted `DERIVED_FROM` edge *rows*,
but a fact's node accumulates one edge per ref *per commit*, so lineage
re-recorded at a second commit double-counted — up to 3x. The badge read 26 where
the connections panel read 13, which voided the design's own claim that the
number is checkable by opening the fact. Now counts distinct target facts; the
fix also dropped the query from 32ms to 11ms.

**The lens never used impact ranking.** The union required unanimity across
mounts, so a single dissenting mount vetoed the axis — and on one real lens the
vetoing mount had zero facts. Now pools fact counts and edge totals across mounts
and applies the same threshold once, through a shared helper so the rule exists
in exactly one place.

## Load-bearing detail

Highlight queries **must** join through `branch_facts`. The graph is temporal and
holds a node per fact *version*; reading `node_props_text` directly resurfaces
versions a prior `review` already subsumed. `TestHighlights_OnlyTheRequestedBranch`
seeds two branches at the same path and asserts each sees only its own version,
with its own edge count.

Navigation is deliberately **live**, not pinned.
`kb/invariants/ui/navigation/every-hop-is-path-plus-commit` governs *following a
recorded reference*; it says in terms that "opening a fact from the list" is
live. Highlights are a listing, so `Highlight` carries no `commit_hash` — its
absence is the design.

Impact is global out-degree: never filtered by `path`, so a fact reports the same
number from the repo root and from its own folder. `default_axis` is repo-scoped
for the same reason — per-folder ratios are noisy and would flip the control
mid-navigation.

## Verification

`go build`, `go vet`, `go test ./...` clean. 711/711 client tests, `tsc` clean.

Verified against a copy of the production `core` KB (1300 live facts, 2721
edges): highlights capped at 10, no observations or references present, impact
descending, confidence visibly unsorted beneath it, axis parameter changes
ordering, path scoping works while impact stays constant, garbage axis falls
back, empty cases return `[]`/`{}` never null.

A recurring defect on this branch was tests that pass for the wrong reason —
fixtures where the assertion could succeed through an unintended path. Several
were found and fixed; each fix was confirmed by mutation (break the behaviour,
watch the test fail, restore).

## Known follow-ups

Not blocking, recorded rather than silently dropped: several lens test fixtures
set an axis with zero counters and so exercise a branch they weren't written for
(their real assertions still hold); `/domains` now pays for the highlights
queries it discards; rows carry no type badge; impact is HEAD-anchored by
deliberate choice, documented as a departure from
`kb/principles/philosophy/historical-not-current`.


knomit:
### Code review

Found 1 issue:

1. Two literal NUL bytes (`0x00`) are embedded directly in the `axisScope` template literal instead of the `\0` escape sequence. This makes git and grep treat the file as binary (`file web/src/RightPanel.tsx` reports `data`, and `grep axisScope` on it silently matches nothing), which breaks search and diff tooling. The line is new in this PR — the base blob has zero NUL bytes. The codebase's existing convention for the same composite-key idea uses the escape: `web/src/Library.tsx` builds `filtersKey` and `reposKey` with `.join('\0')`.

https://github.com/knomit/knomit/blob/b29c900383cdd463e31a4ca06a76a0342033536f/web/src/RightPanel.tsx#L697-L699

PR #63, opened 2026-08-04
fix(llm): the JSON grammar is what breaks a reasoning model's JSON

knomit:
## The bug

`internal/synthesize` asks for JSON with `ForceJSON: true`, which the ollama adapter turned into `format: "json"`. That flag is a decoding constraint the runner has to suspend for the reasoning phase and resume for the answer — and on a reasoning model, that seam is where the output breaks.

Measured against `gemma4:26b-mlx`, running each response through `internal/synthesize`'s own `extractJSON`:

| request | parses |
|---|---|
| `think:true` + `format:"json"` (**before**) | **1/4** |
| `think:true`, no `format` (**after**) | **4/4** |
| `think:true` + `format:<schema>` | 2/4 |

The failures were not fenced JSON, which `extractJSON` already handles. They were text it cannot recover: one response began with a bare literal `json` before the fence, another came back as YAML-ish key/value pairs. The flag whose job was to guarantee JSON was the reason there wasn't any.

## The fix

`ForceJSON` now states a requirement instead of naming a mechanism, and picks how to meet it from the capabilities `/api/show` reports at construction:

- **thinking model** → requirement moves into the prompt, no grammar constraint
- **non-thinking model** → keeps `format:"json"`, which is a hard guarantee where a prompt is only a strong convention, and which these models never had trouble with

A failed probe is not fatal. It only selects between two working strategies, so an unanswered one falls back to the grammar — the behaviour before this change.

## Two further failures, both routine only on a reasoning model

**`num_predict` was `defaultMaxTokens`** — a constant sized to bound spend on a metered API. Reasoning tokens are generated tokens, and ollama counts them against that budget before any reach us, so the budget went to thinking and the answer was never written. At `num_predict=2000`, gemma4 returned `done_reason: "length"` with **empty content**; the same prompt uncapped answered correctly in 7565 tokens. Nothing is metered locally, so the cap bought nothing and truncated real work.

**A truncated stream is a *successful* HTTP response**, and `done_reason` went unread. `Complete` returned whatever it had accumulated — with a reasoning model, routinely `""` — alongside a `nil` error, and `pipeline.go` checks only `err`. An empty string that reads as "the model said nothing" is worse than an error, because nothing downstream can tell it from a real answer. Truncation now returns one.

## Verification

- 7 new tests in `internal/llm/ollama_test.go`, each watched to fail first. Two are controls that must **not** change: the non-thinking model keeps its grammar constraint, and reasoning deltas never reach the caller's `onChunk`.
- End-to-end against a live ollama with `gemma4:26b-mlx`: probe correctly reports `thinking: true`, and `ForceJSON` completions returned parseable JSON 2/2 (24s, 11s).
- Full suite green (`go test ./...`, 0 failures).

## Notes for the reviewer

- `ollamaTestServer` in `resilience_test.go` now delegates to a capability-aware helper. Without it the new `/api/show` probe fell through to the chat handler and was counted as a completion attempt, breaking the retry-count assertions.
- `resilience.go:256` cites `input length … exceeds context window` as "the most common ollama 400". That is no longer true on ollama 0.32.5, which auto-sizes context — a 20k-token prompt returned 200 with `prompt_eval_count: 20023`. The anchored regex is still worth keeping as defence; only the comment's premise is stale. Left alone here to keep this diff to one concern.
- Not addressed, deliberately: `AttemptTimeout` for ollama. Uncapped reasoning makes the 5-minute per-attempt budget the binding constraint, and a hard prompt can approach it (266s measured on an adversarial riddle; 102–173s on realistic synthesis prompts). Worth a follow-up, along with whether a timed-out local reasoning call is worth retrying at all.

### DIFF
--- a/internal/llm/adapter.go
+++ b/internal/llm/adapter.go
@@ -39,7 +39,13 @@ type Message struct {
 
 // CompletionOptions controls provider-specific behaviour for a single call.
 type CompletionOptions struct {
-	ForceJSON bool // when true, constrain output to valid JSON (e.g. Ollama format:"json")
+	// ForceJSON asks for a response the caller can parse as JSON. It states the
+	// requirement, deliberately not the mechanism: a provider is free to meet it
+	// with a decoding constraint, a prompt instruction, or both, and at least one
+	// of those is unavailable on some models. Ollama, for instance, cannot use
+	// its format:"json" grammar on a reasoning model without corrupting the
+	// answer (see ollama.go's jsonStrategy), so it asks in the prompt instead.
+	ForceJSON bool
 }
 
 // LLMAdapter is the common interface implemented by every provider.
--- a/internal/llm/ollama.go
+++ b/internal/llm/ollama.go
@@ -28,10 +28,13 @@ type OllamaAdapter struct {
 	model  string
 	client *http.Client
 	// thinks records whether the server reports this model as capable of
-	// reasoning. It decides how ForceJSON is honoured, so it is resolved once at
-	// construction rather than per request: the answer is a property of the
-	// model, and asking on every completion would put an extra round trip in
-	// front of work that is already the slowest thing in the pipeline.
+	// reasoning. It gates two things, not one: whether think is sent at all
+	// (every completion, ForceJSON or not) and how ForceJSON is honoured when it
+	// is asked for. It is resolved once at construction rather than per request
+	// because the answer is a property of the model, and asking on every
+	// completion would put an extra round trip in front of work that is already
+	// the slowest thing in the pipeline. The cost of getting it wrong is on
+	// reportsThinking.
 	thinks bool
 }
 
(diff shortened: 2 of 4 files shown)


knomit:
### Code review

Found 1 issue:

1. `reportsThinking`'s doc comment understates what a failed probe costs. It says the failure is safe because "The probe only selects between two working ways of asking for JSON" — but `a.thinks` also gates `think: true` on every completion, independent of `ForceJSON`. A probe that fails (or returns non-200, or fails to decode) silently disables reasoning mode for the adapter's entire lifetime, and puts `format:"json"` back on a reasoning model — the exact combination this PR measured at 1/4 parseable. Worth either widening the comment or making the fallback narrower.

https://github.com/knomit/knomit/blob/829666ddedb53df8c36bf18808824f5d864b70e0/internal/llm/ollama.go#L89-L98

The unconditional `think = true` that the same flag drives:

https://github.com/knomit/knomit/blob/829666ddedb53df8c36bf18808824f5d864b70e0/internal/llm/ollama.go#L203-L208

PR #65, opened 2026-08-05
control.db becomes the control plane: authoritative repo registry, and the origin credentials that make its recovery real

knomit:
**This PR now carries the whole control.db migration.** It was previously stacked on #61; that PR has been closed and its 17 commits are all here, unchanged, as the first half of this branch. Nothing was squashed or rewritten — #61's review history stays readable on #61 itself.

The branch has been merged with `dev` (see the final merge commit for the conflict resolutions, including two silent ones the textual merge could not see).

Read it as two movements: **Part 1** makes `control.db` authoritative for *what repos exist*, so a repo whose `.db` is lost is re-cloned at boot from its recorded origin. **Part 2** makes that promise true, by giving the control plane the credential the re-clone actually needs.

---

# Part 1 — The repo registry (was #61)

## The problem

**The filesystem was the registry.** `Manager.Start` globbed `repos/*.db` and opened whatever it found. Three consequences:

1. **A repo's origin lived only inside its own database.** Lose the file and nothing anywhere records where it came from. The one thing that could rebuild it is the one thing that went missing with it.
2. **Archive metadata lived in `repos/archive/<id>.json`** — a sidecar next to the SQLite file, with its own parse-failure and partial-write modes, describing control-plane state that had no business being outside the control plane.
3. **One repo was privileged.** `core` was created implicitly at boot by `initDefaultGit`, outside every registration path, driven by a *server-wide* `git.origin`. That produced asymmetries that had nothing to do with defaults: a cloned repo on a server without `git.origin` never index-synced its upstream and never reconciled at boot, because both were gated on the server-wide key rather than the repo's own remote.

A new `repos` table in `control.db` answers "what should exist, and where did each come from?" — the question an empty disk cannot answer. Every repo then arrives through one path, which is what makes recovery uniform.

## `control.db` gets versioned first

It had **three tenants creating their own schema at open time**: a `lensSchema` const, a `repoSettingsSchema` const, and a runtime `ALTER TABLE lenses ADD COLUMN description` that swallowed `duplicate column name`. Adding a fourth tenant to that was not on.

`migrate.Control` is a separate migration set from `All`/`Core` — different file, its own `schema_migrations`, stock `sqlite3` driver since `control.db` holds no vectors. Migration 1 is the pre-`description` schema with `IF NOT EXISTS`, so an install created before versioning is **adopted** rather than failed. Migration 2 is that `ALTER`; where the column already exists, the existing `upWithRecovery`/`alreadyApplied` path marks it applied — the versioned equivalent of the error the runtime code used to swallow. Verified against a real pre-migration `control.db`; see **Rehearsals** below.

## The registry

`repos(name, origin_url, origin_branch, state, archive_id, created_at, archived_at)`, primary key **`(name, archive_id)`**.

The composite key is not tidiness. *A repo name is unique only among ACTIVE repos.* Archiving `work` and creating a new `work` is supported and tested, as is archiving the same name twice — so with `name` alone, the new active row overwrote the archived one and the archived repo became unrecoverable the moment its name was reused. Active rows carry `archive_id = ''`, so a name still has at most one active row and upserting an active repo by name still updates in place.

Three API decisions, each with a defect behind it:

- **`EnsureActive` is not a convenience wrapper over `Upsert`.** `Upsert` is a whole-row write — every column comes from `excluded` — so upserting a partially-filled record **blanks the origin and restamps `created_at`**. Right for a repo being created, wrong for one being *re*-registered. `Rescan` adopts a database that may already have a row (`Start` skips-and-logs a repo whose open failed, leaving its row behind), and blanking `origin_url` there destroys the one field the rebuild path reads. `EnsureActive`'s conflict arm touches `state` and `archived_at` only.
- **There is deliberately no `SetState(name, state)` and no `Delete(name)`.** Both read as if a name identified one repo, which stopped being true at the composite key. `SetState("work", archived)` would flip every row sharing the name and leave the active one with an empty `archive_id` — an archived row no `Restore` or `Purge` could ever address. Archiving is "insert the archived row, retire the active one": `DeleteActive` + `DeleteArchive`. The comment saying so is load-bearing.
- **`DeleteArchive` is keyed by id, never by name** — the archived repo's name may since have been claimed by a live active repo.

## Startup

`adoptFromFilesystem` runs **before anything opens**, and only when the registry is empty: on an upgrade the disk is the only record of what exists, so the filesystem gets one last say. Afterwards the registry is authoritative and the disk is never consulted again. It folds `repos/*.db` into active rows and `repos/archive/*.json` manifests into archived ones.

`Start` then reconciles each active row against the disk: **present** → open; **absent with a recorded origin** → re-clone through the ordinary create path (so the clone inherits `setupIndex`'s both-branch sync rather than reimplementing it); **absent with nothing to rebuild from** → logged and skipped.

**No single repo can fail the boot.** Every per-row outcome is logged at ERROR and skipped — an open that failed, a clone that failed, nothing to rebuild from. One repo's unreachable origin or corrupt database must not take the other repos on the instance offline with it. The row survives in every case, so the next restart retries and the state stays diagnosable. `Start` returns an error only for whole-manager conditions: the repos directory, `control.db`, the session reaper's config.

`isValidRepoName` is now applied on the **read** paths too (the reconcile, `Rescan`, adoption), so a hand-edited or corrupted row skips-and-logs instead of refusing the boot.

## Origin write-through

`RecordOrigin` copies a repo's live origin from its store into its registry row. The store's `remotes` table stays the source of truth; the row is a copy for the one case the store cannot serve — the store file is gone. A copy is only worth having if it is kept current, so **every path that changes an origin calls it**: clone-mode `Create`, the boot reconcile, `Rescan`, `PUT`/`PATCH`/`DELETE` on `/repos/{repo}/origin`, and both origin-session commit paths that write the remote directly rather than through the provider. Without the last of those, an origin attached *after* creation lived only in the store — the exact hole the registry exists to close.

It is **authoritative**, which means it CLEARS the row when the store reports no origin. A disconnect is an origin change like any other, and a registry that kept the old URL would have the next boot re-clone from a remote the user deliberately removed.

The subtlety is `originOf`'s third return, and conflating the two empties is a bug in both directions:

- `ok=true, url=""` — the repo **genuinely** has no origin. Bring the registry into line, *including clearing* what it holds.
- `ok=false` — the store could not be READ. Nothing was learned; leave the row exactly as it is. Treat this as "no origin" and a transient read failure erases the one field a rebuild needs.

The `PUT /origin` write-through is placed **before** `ActivateSync`, deliberately: that call can fail the request with a 502 while the origin *is* already persisted, so a write-through after it would be skipped for exactly the origin most likely to need re-cloning later.

## Archive / restore / purge

The registry row **replaces** `repos/archive/<id>.json`. `Archive` inserts the archived row and *then* retires the active one — the window where both exist is recoverable, the window where neither does is not. `Restore` mirrors it: active row first, then `DeleteArchive`, carrying `CreatedAt` across so a restored repo does not look newly created. `Purge` deletes by archive id. `ListArchived` reads rows; the `ArchiveInfo` shape and newest-first ordering callers see are unchanged. A leftover legacy manifest is removed on restore/purge — left in place, a later adoption on a machine that lost `control.db` would resurrect a dead entry pointing at a database that is gone.

**Fixed here:** an archive rollback could leave a repo **live but unregistered**. The rollback restored the captured active row only `if hadPrior`; when the capture found nothing or failed, nothing restored a row at all. Since `Start` reads the registry and no longer globs, the repo served normally until the next restart and then simply was not there, with nothing in the log connecting the two — and `RecordOrigin` could not repair it either, since it declines to write when the row is absent. The `!hadPrior` arm now calls `EnsureActive`, and the three recovery sites collapse into `reinstateLive`, which re-derives the origin from the store now the repo is open again. The regression test injects the fault in SQL — a trigger rejecting inserts of *archived* rows only, so recording the archive fails while the rollback's own insert still works. Confirmed failing without the fix.

## The default repo is gone

This is forced, not opportunistic: `core` was the last path by which a repo could exist without a registry row.

**Deleted:** `config.DefaultRepoName`; `GitConfig.Origin` and its `KNOMIT_GIT_ORIGIN` overlay (only ever applied to `core`, gated on `isDefault`); `repoBuilder.isDefault` and every branch on it; `initDefaultGit`; `resolveOriginUpstream`; `ErrCannotArchiveDefault`; `ErrCannotArchiveLast`.

**Three things moved rather than vanished**, because deleting them outright would have silently dropped behaviour:

- **Remote-HEAD detection.** It lived in `initDefaultGit`; the clone path recorded a hardcoded `"main"` when no branch was requested, so a master-default origin got refspecs pointing at a branch that does not exist. `detectUpstream` now runs in `initClone` *before* the clone, so clone, refspecs and the `remotes` row all name one branch.
- **The ontology seed.** `initClone` passed no seed files. That only matters for an *empty* origin — a non-empty one brings its own — but a repo cloned from an empty remote had no ontology at all.
- **`upstreamMain` and the startup reconcile.** Both were gated on the server-wide origin. `resolveUpstreamMain` reads the repo's own stored remote; empty means no origin. `warnIfUpstreamMissing` is a loud diagnostic and deliberately not a self-heal — re-detecting every boot costs a network round-trip per repo and could overrule a deliberate choice.

The prefer-`main` rule also moves out of `InitFromRemote`'s guard into `detectFromRemote`. Once `initClone` resolves the branch itself and passes it in **non-empty**, that guard is skipped entirely — so a remote whose symbolic HEAD points at an agent branch (its GitHub default set to `agent/<host>`) would otherwise make that agent branch the consensus upstream.

**Archiving the last repo now succeeds** and leaves zero repos. A user who created a repo by mistake has to be able to undo it, and `Restore` brings it back.

**`knomit reset` is removed.** It deleted a repo's `.db` directly, which predated the registry; once the registry became authoritative the row survived the delete and the next boot either re-cloned the repo straight back — reset as a no-op with extra steps — or logged it forever as registered-but-unrebuildable. Removal goes through the registry now: `DELETE /api/v1/repos/{repo}` to archive, `DELETE /api/v1/archived/{id}` to purge. `app.ResetRepo` had no other caller and goes with it. `root_test.go` pins the command set **exactly**, so a command added without a decision behind it is visible in the same place.

`knomit verify` and `knomit-bridge` now require `--repo` (or `--all`), validated *before* `app.New` so a missing flag is not buried behind seconds of boot. `verify --all` with no repos says so rather than exiting 0 in silence, which read as "everything verified clean".

## Zero repos is a first-class state

A fresh install has none until the user makes one, so the empty screen must be a way **in**: it renders the repo manager alongside it and creates the first repo right there, instead of telling the user to go run a command line and reload. `RepoManager` lands on the create form when there is no current repo — via a `homeView` rather than `{kind:'repo', name:''}`, which is a non-null selection that slips past the `sel ??` fallback and renders a detail pane fetching `/api/v1/repos//origin`.

---

# Part 2 — Origin credentials in control.db

## The problem

Part 1 records `origin_url` and `origin_branch`. It does not record the credential — that lives encrypted in the repo's own `remotes` table, **inside the very file whose loss triggers the rebuild**. So the registry row promises a recovery it cannot perform: boot retries the clone forever, failing authentication each time.

The code conceded it in its own error text:

> "a private origin using token auth can never be re-cloned this way: the token was stored encrypted inside the missing database, so it is gone with it"

`git` is the source of truth for facts, but reaching git needs credentials. So the control plane has to be sufficient to re-create anything it claims exists.

Unaffected: SSH origins (the key is a separate file), public HTTPS, `file://`, and repos with no origin. The gap is specifically **token and basic auth**.

## Ownership

`control.db`'s `repos` row owns origin **intent** — `origin_url`, `origin_branch`, `auth_method`, `auth_token` (AES-256-GCM, keyed from the agent SSH key via HKDF, so this is *control.db + the agent key*, never control.db alone). The store's `remotes` row demotes to materialized runtime state: url, branch, intervals, and the `last_*` status columns, with its auth columns **empty from here on**. Status genuinely belongs to the local repo; intent does not.

Writes funnel through `Manager.SetOrigin`/`ClearOrigin`, replacing four scattered `SetRemote` sites and one `DeleteRemote`. **control.db is written first, always** — a crash between the steps must leave the credential recorded and the store unwired, never the reverse, because the reverse is the bug this PR removes. Both orderings are pinned by tests that fail when the steps are swapped.

## Two guardrails, both learned from bugs

**The token is never a field on `RepoRecord`.** `ListArchived` copies that struct into `ArchiveInfo`, which is serialized to `GET /api/v1/archived` — a token field there is one plausible-looking field copy away from publishing every credential over HTTP, and such a line reviews as ordinary. It is reachable only through `OriginCredential`. A byte-level test asserts no credential appears in `/repos` or `/archived`.

**`Upsert` writes neither auth column.** It builds its INSERT from a `RepoRecord`, which structurally cannot carry a token — so it would pass `''`. Naming `auth_token` in the `ON CONFLICT` arm would blank the credential on every `RecordOrigin` write-through: every boot, create, rescan, and origin edit. Immediate, total, nothing logged. Columns left unnamed keep their value.

## The boot repair that had to be built

The design originally claimed a crash mid-write would be "repaired by the next boot materializing." **No such step existed.** What actually ran was `Start` → `RecordOrigin`, reading the origin *from the store* and writing it into control.db — the opposite direction. Both crash states were made worse, not repaired.

`reconcileOrigin` implements the missing step, and it cannot be a plain inversion: `adoptFromFilesystem` writes rows with an empty `OriginURL` (it globs `.db` files and never opens them), so a blank row means *either* "the user disconnected it" *or* "control.db has not learned it yet." Unwiring blank rows would strip a working origin off every pre-registry repo. The rule resolves that by severity — push down when control.db knows an origin, pull up when it does not, and leave a blank branch blank because it self-corrects through `detectUpstream` on rebuild while a branch adopted at boot cannot be told apart from `SetRemote`'s hardcoded `"main"` substitution.

**Documented residual:** a crash midway through `ClearOrigin` can still re-record a disconnected origin. That is deliberate — resurrection costs a second disconnect, unwiring costs a working remote — and `ClearOrigin` now returns a failed unwire so the incomplete disconnect is visible rather than reported as 204.

## Migration

Existing installs are migrated at boot. **A non-empty `auth_token` in the store is the unmigrated signal** — no bookkeeping column that could itself drift. Migration empties it, so the check is self-describing and idempotent.

It must not use `GetRemote`: on a decrypt failure that returns the *stored bytes* as though they were plaintext, and encrypting those would produce a credential that looks migrated, decrypts to ciphertext, and can never authenticate. `LegacyAuth` is the strict reader that distinguishes "nothing stored" from "stored but undecryptable."

A repo whose credential cannot be moved is **not served** — its registry row survives so the state stays diagnosable and the next boot retries. The blast radius is deliberately narrow: only repos that *actually hold* a credential can be refused, so SSH, public, and origin-less repos are never blocked by an unreadable agent key. The gate guards all three doors into service (`Start`, `Rescan`, `Restore`); `Create` and `reinstateLive` are ungated and the comment now names both and says why each is safe.

## Verified end to end

`test/storytests/control_db_rebuild_test.go` drives a bare git origin behind **HTTP basic auth**, because a `file://` origin proves nothing — go-git's file transport never authenticates, and the earlier `file://` rebuild test passes with the credential-copy code completely removed. The gate test fails with `authentication required: unauthorized` when that code is stripped, while the unauthenticated repo still returns.

Also verified by hand on an isolated home and a spare port: created a token-auth clone, confirmed `origin_branch=master` was genuinely detected rather than defaulted, deleted every `repos/*.db`, rebooted, and the repo came back with its facts and **no re-authentication**. No plaintext credential anywhere in `control.db`; the repo's own store held none. The origin-less repo was correctly absent from the API with its registry row intact and an ERROR naming it.

---

# Merging `dev`

Four textual conflicts, all in `internal/repos`, plus **two silent semantic ones** the textual merge could not see:

- **`builder.go`** — dev edited `initDefaultGit` to seed the ontology at its new `.domains` path; this branch deleted `initDefaultGit` and `resolveOriginUpstream` outright with the default repo. Kept the deletion: there is no default repo to bootstrap, so dev's edit had no subject. `loadOntology`'s canonical/legacy preference merged cleanly and is kept in full.
- **`lifecycle.go` (silent)** — `initClone`'s empty-origin seed is code only this branch has, so git merged it without a conflict while it went on writing the **legacy** `domains/ontology.yaml`. Repointed at `OntologyPath`.
- **`builder_ontology_test.go`, `builder_adopt_test.go`, `manifest_test.go`** — dev's fixtures reach for `config.DefaultRepoName`, which this branch removed. Switched to `testRepoName`.
- **`builder_ontology_test.go` (silent)** — `bootKnomitWithLegacyOnlyOntology` and `bootKnomitWithBothOntologies` are new on dev and arrived whole, so they assumed a repo exists after `Start()`. It does not: `Start` opens only what the registry lists. Both now create the repo explicitly, which is what made their three tests fail rather than pass vacuously.

`TestLoadOntology_PrefersDotDomains` was re-verified by inverting `loadOntology`'s path preference and confirming it goes red, since its fixture changed here.

## Verification

- `CGO_ENABLED=1 go test ./...` — no failures
- `CGO_ENABLED=1 go vet ./...` — clean
- `cd web && npm run build` — clean
- `cd web && npx vitest run` — 721/721

## What was dropped from #49 on the way

- `Deps.StrictMissing` / `ErrRepoUnrecoverable`. It made `Start` refuse the boot for a repo with no database and no origin, and existed purely for the replica. Without a replica that reasoning does not survive, and what is left is a flag that turns one repo's bad day into a whole-instance outage. The rule is now unconditional skip-and-log.
- `Manager.OpenDBPaths` / `ArchivedDBPaths` — they existed only to feed the tracker.
- Reserving the repo name `control`. Its three stated consequences are all replica-namespace consequences; `<home>/control.db` and `<home>/repos/control.db` do not collide on disk. Worth doing on hygiene grounds, but as its own change.
- The swapstore staging / WAL-set-aside durability fixes. Real and independent of both backup and the registry; they belong in their own PR.
- The `reserveNameAndOrigin` wrap around `Archive`. It closes a genuine Create-during-Archive race on the `.db` file, but #49 justifies it entirely in tracker terms.

## Rehearsals

Run by hand against a **real pre-migration `KNOMIT_HOME`** — a `control.db` carrying `lenses` (with `description`), `lens_reads` and `repo_settings`, and no `schema_migrations`:

- **Upgrade.** Migrations 1–4 apply; migration 2 hits its own committed body and completes through `upWithRecovery`; the repo is adopted and verifies CLEAN; a stray `core 1.db` is skipped for an invalid name; a legacy archive manifest becomes an archived row with its origin and `archivedAt` intact; 3 lenses and the settings table survive untouched.
- **Idempotence.** Second boot logs no adoption at all and leaves every row unchanged.
- **Boot resilience.** A row pointing at an unreachable origin with no database, plus a row with neither — the server starts, serves the healthy repo, logs both at ERROR with the recovery spelled out, and leaves both rows in place.

## Follow-ups, not in this PR

Two **pre-existing** bugs surfaced during the manual run, both confirmed present at the merge base:

- A clone-mode create starts two reconcile loops; `ActivateSync` cancels the first, and its dying fetch persists `last_status="error"` over a healthy repo until the next tick — a red remote-error banner in the UI after every clone-mode create.
- `handlers_repos_create.go` passes `r.Context()` into `Create`, tying create-time work to the request lifetime.

Also deferred: `CopyCredential` and `SetOriginCredential` both ignore `RowsAffected`, and `EnsureActive` has no guardrail test.

KB facts to write once this lands: that `control.db`'s `repos` table is authoritative and nothing may reintroduce a `repos/*.db` glob as a source of truth (`adoptFromFilesystem` being the sole, registry-empty exception), and that every origin-changing path must call `RecordOrigin`.


knomit:
## CI caught a real hole, not a stale test

The first run failed one contract cell — `TestContract_AuthResolutionFailure_SurfacesVisibleError`. Two commits fix it, and the first is a production change.

**Why it escaped local runs.** These cells are `//go:build contract`, and CI runs them separately (`tests.yml:91`: `go test -tags contract ./test/storytests/`). Every test command used while developing this branch omitted the tag, so this whole class of test was invisible throughout — including to the per-task and whole-branch reviews. My fault, and worth knowing for anything else that touches auth resolution.

**The hole.** `migrateCredential` returned early when the store's `auth_token` was empty, so a legacy row carrying an `auth_method` with **no token** never had its *method* carried into control.db. Since auth now resolves from control.db only, that method was silently dropped — and for a URL `resolveAuthWithOrigin` cannot re-infer `ssh` from (anything not `git@`/`ssh://`), resolution fell through to **anonymous**. A pre-upgrade install with `auth_method='ssh'` on an `http://` origin used to fail loudly with "ssh auth requires a key path"; on this branch it would sync anonymously and report `last_status="ok"`. That is precisely the silent-anonymous downgrade the cell exists to forbid.

`20c002da` carries the method regardless of the token. A method-only row holds no secret, so it needs no `Crypt` and can never cause the gate to refuse a repo — a failed carry logs at ERROR and the repo still serves. The token-bearing refuse-to-serve path is untouched, and `ClearAuth` still runs only after a successful carry.

**The cell got stronger, not weaker.** `b14807cf` keeps its original store-side `UPDATE remotes SET auth_method='ssh'` and adds a `Restart()`, so the boot migration is what carries the method into control.db. All four assertions are byte-identical — only the setup gained a precondition. So the cell now covers the migration *and* the contract, and it was verified red without `20c002da`.

Green now: `go test ./...` (43 packages), `go test -tags contract ./test/storytests/`, and `go test -tags desktop ./tools/desktop/...`.

knomit:
Will not merge, stale implementation

PR #68, opened 2026-08-05
feat: KB repo layout — README.md, LICENSE, and private dot-paths

knomit:
Reshapes a KB repo's on-disk layout so it reads as an ordinary git repository to an ordinary git host.

**There is no migration — existing KB repos are renamed by hand.** Each change below states what an unmigrated repo does.

## Changes

**`kb.md` → `README.md`** — every major git provider already renders `README.md` at a repo root as the repository description, which is exactly what this file is. Clean break: nothing reads `kb.md`. An unmigrated repo shows an empty description card until the file is renamed.

**`LICENSE` at the repo root** — read and exposed on the single-repo GET, rendered as preformatted text (a markdown renderer reflows a licence's newlines away). Read-only by design: no write path, no seeding, no OKF export.

**Dot-prefixed segments are private** — `fact.IsPrivatePath` excludes them from fact *discovery* at any depth, including under the ontology root, and the fact-creation paths refuse to allocate one. `kb/.drafts/` is therefore a usable private stash.

Private governs **walking, not opening** — knomit still reads `.domains/ontology.yaml` by name. That is why the check is deliberately *not* in `store.validatePath`, which also guards `README.md` and the ontology write.

**`domains/` → `.domains/`** — the ontology is configuration, not knowledge. Unlike the README this *does* keep a legacy read fallback, because the cost of a miss differs: a missing README costs a description, a missing ontology falls back to `DefaultOntology()` and would silently validate new facts against the wrong taxonomy. `loadOntology` remembers which path it read and the preset refresh writes back to that same path — otherwise a legacy repo grows a second ontology file with the stale one left as a decoy.

**`spec/mbekg.md`** — §3.8 asserted "Exclusion is parse-failure-based, not path-based. There is no list of excluded paths." That is now false, and had already drifted from `isFactPath` before this work. Rewritten to the location-based contract, and its corollary that a stray file with valid frontmatter is a fact *wherever it sits* is deleted — that is the exact misreading this change eliminates. Also fixes a pre-existing error in the §3.1 diagram, which placed the root manifest inside `kb/` rather than at the tree root.

## Note on `WriteRootFile`

`factIndex.writeFile` lowercases every path — a fact-path rule that prevents case-duplicate topics (`AI` vs `ai`). It never mattered for `kb.md`, but would have committed `readme.md`. `store.WriteRootFile` is the case-preserving door for non-fact root files, restricted to root level so it cannot become a bypass of fact-path normalization.

## Review

Eight tasks, each independently reviewed, plus a whole-branch review. Six defects were caught and fixed before merge, including two that this branch itself introduced:

- A **fourth** fact-creation site — the `PUT .../facts/{path...}` endpoint, which takes a fully caller-supplied path. This branch had silently changed it from "accepted and indexed" to "committed and permanently invisible, returns 200".
- Three ListDir-backed topic browsers that listed `.drafts` as an ordinary topic, contradicting the §3.8 MUST shipped in this same PR.

Three of the six were tests that passed under a broken implementation; each was replaced with one demonstrated to fail against the specific regression it guards.

## Verification

- `CGO_ENABLED=1 go test ./...` — 43 packages, 0 failures
- `CGO_ENABLED=1 go vet ./...` — clean
- `cd web && npm run build` — clean (the real `tsc -b`; `npx tsc --noEmit` type-checks zero files in this repo)
- `cd web && npx vitest run` — 718/718

### DIFF
--- a/internal/fact/private.go
+++ b/internal/fact/private.go
@@ -1,0 +1,27 @@
+package fact
+
+import "strings"
+
+// IsPrivatePath reports whether any segment of path begins with ".".
+//
+// A dot-prefixed directory or file is MACHINERY, not knowledge: .github/ holds
+// CI config, .domains/ holds the ontology definition. Private paths are
+// excluded from fact DISCOVERY everywhere — the search indexer, Verify, and the
+// OKF exporter all skip them, and the fact-creation paths refuse to allocate
+// one.
+//
+// Private means "not knowledge content", NOT "invisible to knomit". knomit
+// still reads specific known paths by name, which is precisely what lets it
+// load .domains/ontology.yaml. The rule governs walking, never opening.
+//
+// This is a LOCATION test, so it extends the ontology-root scope rule rather
+// than competing with it: membership is decided by where a file sits, never by
+// whether its bytes happen to parse.
+func IsPrivatePath(path string) bool {
+	for _, seg := range strings.Split(path, "/") {
+		if strings.HasPrefix(seg, ".") {
+			return true
+		}
+	}
+	return false
+}
--- a/internal/fact/private_test.go
+++ b/internal/fact/private_test.go
@@ -1,0 +1,31 @@
+package fact
+
+import "testing"
+
+func TestIsPrivatePath(t *testing.T) {
+	cases := []struct {
+		name string
+		path string
+		want bool
+	}{
+		{"ordinary fact", "kb/architecture/store/a1b2c3d4.md", false},
+		{"leading segment", ".github/workflows/ci.yml", true},
+		{"ontology dir", ".domains/ontology.yaml", true},
+		{"middle segment", "kb/.drafts/a1b2c3d4.md", true},
+		{"filename segment", "kb/architecture/.wip.md", true},
+		{"deep middle segment", "kb/a/b/.c/d.md", true},
+		{"root manifest", "README.md", false},
+		{"licence", "LICENSE", false},
+		{"dot inside a segment", "kb/architecture/v1.2/a1b2c3d4.md", false},
+		{"parent traversal", "kb/../secrets.md", true},
+		{"current dir", "./kb/a/b/c.md", true},
+		{"empty", "", false},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			if got := IsPrivatePath(tc.path); got != tc.want {
+				t.Errorf("IsPrivatePath(%q) = %v, want %v", tc.path, got, tc.want)
+			}
+		})
+	}
+}
--- a/internal/mcp/learn.go
+++ b/internal/mcp/learn.go
@@ -216,6 +216,13 @@ func validateAndBuildFacts(ontology *fact.Ontology, ontologyRoot string, inputs 
 		// Build path with server-generated UUID.
 		path := fact.BuildFactPath(ontologyRoot, fi.Topic, fi.Category)
 
+		// See handlers_fact_create.go: a private path is written and then
+		// ignored by every reader, so it is refused rather than accepted.
+		if fact.IsPrivatePath(path) {
+			return nil, nil, nil, nil, fmt.Errorf(
+				"fact %d: topic/category resolves to %s; a path segment beginning with '.' is private and cannot hold a fact", i, path)
+		}
+
 		domain := fi.Domain
 		if domain == nil {
 			domain = []string{}
--- a/internal/mcp/learn_paths_test.go
+++ b/internal/mcp/learn_paths_test.go
@@ -81,3 +81,27 @@ func TestMergeFacts_LineageRefUsesRawPath(t *testing.T) {
 	// Identity stays normalized — unchanged from the pre-refactor behaviour.
 	require.Equal(t, strings.ToLower(rawPath), merged.Path())
 }
+
+// knomit_learn must refuse exactly what the REST create path refuses: the two
+// share BuildFactPath, so they must share the rule that bounds its output.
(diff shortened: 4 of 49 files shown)


knomit:
## 📌 Post-merge follow-ups for knomit (do these when this lands, not before)

Recording facts about unmerged code is how a corpus rots, so these are deliberately deferred. A future session should run them **after** this merges to `dev`.

Everything below is stated so it can be executed without the original session's context.

---

### 1. Update `kb/invariants/store/indexing/ontology-root-scope/1b35d56b.md`

`/knomit-update kb/invariants/store/indexing/ontology-root-scope/1b35d56b.md`

Two reasons it has drifted:
- Its rationale names **`kb.md`** as the motivating example of a root file that would parse as a fact. That file is now `README.md`.
- Its rule is **extended**, not replaced: index membership is still decided by LOCATION, but location now also excludes any dot-prefixed path segment at any depth. Both `entities` and body need the update.

The invariant's core claim is unchanged and still correct — do not retract it.

### 2. Update `kb/decisions/repos/description-from-kbmd/8992c2a9.md`

`/knomit-update kb/decisions/repos/description-from-kbmd/8992c2a9.md`

Named for the renamed file. Its `kb.md` entity, its title, and its body all need updating to `README.md`. The decision itself (repo description = verbatim manifest read at HEAD, single-repo GET only, collapsible in the detail pane) still holds — it now also has a `LICENSE` sibling on the same endpoint.

The slug `description-from-kbmd` will be wrong after this. Judgement call whether that's worth a retract-and-rewrite or whether an updated body is enough.

### 3. Write a new invariant for the private-path rule

This is the important one, and **the counts below are the whole reason it's worth writing.** The original implementation plan claimed three creation sites and four discovery walkers. Both were wrong, and the gaps were only caught in the final whole-branch review — after seven of eight tasks had already shipped.

Verified counts at `c4223827`:

**4 fact-creation sites (reject on write):**
| Site | Note |
|---|---|
| `internal/web/handlers_fact_create.go:79` | `POST`, path derived from `domain[0]` |
| `internal/web/handlers_fact_write.go:161` | **`PUT`, fully caller-supplied path** — the one the plan missed |
| `internal/mcp/learn.go:221` | `knomit_learn` |
| `internal/synthesize/decision.go:49` | inside `validateOutputPath`, which has **4** callers: `decision.go:152` merge, `:264` distill, `:447` propose, `discovery.go:260` emergent |

**7 discovery guard sites (skip on walk):**
| Site | Note |
|---|---|
| `internal/store/factpath.go:48` | in `isFactPath` — one call site, but covers the search indexer **and both `verify.go` sites** |
| `internal/okf/source/facts.go:44` | must sit **before** `looksLikeFact`, or private drafts are reported as lost knowledge on every export |
| `internal/okf/source/history.go:458` | `okfChangeFromFile` |
| `internal/okf/source/history.go:511` | `okfDeletionFromFile` |
| `internal/web/handlers_topics.go:133` | ListDir-backed, **not** index-backed |
| `internal/web/handlers_topics.go:261` | ListDir-backed |
| `internal/web/handlers_lenses_topics.go:113` | ListDir-backed |

The fact must carry these three traps, because each one was actually hit during this work:

1. **The check must NOT go in `store.validatePath`.** It looks like the ideal chokepoint — every write reaches it — but it also guards the writes for `README.md` and `.domains/ontology.yaml`, so putting it there breaks the boot-time ontology refresh.
2. **Private means excluded from DISCOVERY, never from ACCESS.** knomit reads `.domains/ontology.yaml` by name, and that is permitted. The rule governs walking, not opening. An implementer who reads it as "never touch dot-paths" builds a reader that cannot load the ontology.
3. **The ListDir-backed browsers are discovery surfaces too.** They walk the raw git tree rather than the index, so they do not inherit `isFactPath`'s guard and must filter independently — and they must test the *constructed full path*, not the bare entry name, or a private ancestor directory slips through.

Suggested placement: `kb/invariants/fact/private-paths/…`, linked to the four decision facts already written under `kb/decisions/*/kb-repo-layout/*` (`readme-clean-break`, `license-read-only`, `private-dot-paths`, `ontology-dot-domains-fallback`).

### 4. Optional — a methodology fact worth considering

Three of the six defects caught in review were **tests that passed under a broken implementation**: an ordering assertion whose fixtures all parsed so the branch it guarded was never reached; a guard with no coverage at all; and a "canonical wins over legacy" test that seeded only the canonical file. All three came from the plan's own test fixtures, which suggests a systematic habit rather than three slips.

The generalisable lesson: **when a test exists to pin a specific regression, prove it fails against that regression before trusting it** — construct the broken implementation, watch the test go red, then restore. Cheap, and it caught all three here once applied.

---

### Already recorded (do not duplicate)

The four design decisions were written to knomit **before** implementation began and need no follow-up:

- `kb/decisions/repos/kb-repo-layout/readme-clean-break/b534400d.md`
- `kb/decisions/repos/kb-repo-layout/license-read-only/2b569af1.md`
- `kb/decisions/fact/kb-repo-layout/private-dot-paths/b3425b43.md`
- `kb/decisions/repos/kb-repo-layout/ontology-dot-domains-fallback/318f619a.md`

Note the third overlaps with item 3 above — check it before writing, and extend it rather than duplicating if it already covers enough. It was written from the design and carries the pre-implementation counts, so at minimum its site counts need correcting.

### Ref format reminder

`src://` refs need the source repo's own components, computed in this checkout at the merge commit:

```bash
git rev-list --max-parents=0 HEAD | cut -c1-12   # repo-id (7b4887ce51d9 for knomit)
git rev-parse HEAD                                # commit, full 40 hex
git rev-parse <commit>:<path>                     # blob, full 40 hex
```

PR #69, opened 2026-08-06
ci(release): distill release changelogs and give stable a what's-new section

knomit:
Fixes two independent defects, one in each release workflow, and gives the appcast feed a description worth reading.

**`release.yml` emitted an unreadable changelog.** The step ran `git log --no-merges` over the whole range, so the current `dev-latest` release carries **73 lines** of raw conventional-commit subjects — mostly intra-PR churn (`fix(review): budget the page by measuring it, not by a percentage of it`, `bump version`). `--no-merges` is what made it maximal: this repo merges via PRs, so the merge commits it discarded were exactly the summary layer. That range contains 9 PRs.

**`release-stable.yml` had no changelog at all.** Its `notes.md` was Downloads / Updates / Install / Verifying and nothing else. `tools/appcast` then set `Notes: rel.Body`, so the `<description>` every macOS client polls was install boilerplate. Nothing was dropping the changes — they were never written.

## `tools/relnotes`

A new command beside `tools/appcast`, with two subcommands that fail in deliberately opposite directions.

**`changes -from X -to Y`** is the changelog and **fails loudly**. It walks merge commits, keeping only subjects matching `^Merge pull request #(\d+) `, and resolves each through `gh pr view`. The regex rather than bare `--merges` is load-bearing: the reference range has 10 merge commits and 9 PRs — `Merge remote-tracking branch 'origin/dev' into …` is a branch sync, not a changelog entry.

Commits reachable from no PR merge are listed separately as direct commits, so a hotfix pushed straight to `dev` can never vanish silently. That path does not fire on the reference range (all 73 commits there arrived through PRs) and is covered by unit test rather than by history.

**`distill`** is the optional summary and is **fail-open**. Absent `GEMINI_API_KEY`, transport error, HTTP error, malformed JSON, empty result, unreadable stdin — every one produces a note on stderr, nothing on stdout, and **exit 0**. A release must never fail because a summarization call failed. It runs under a 60s timeout, because the fail-open contract covers errors but not hangs, and the SDK sets no default deadline.

`changes -bodies-out <path>` writes a body-enriched variant from the same `Collect` result, at no extra API calls, so the model sees each PR's rationale while stdout keeps carrying the title-only changelog the release notes are built from.

## The appcast fence

Release notes now mark their summary region with `<!-- appcast:begin -->` / `<!-- appcast:end -->`, and `ItemsFromReleases` carries only that into `<description>`.

**A body with no fence returns whole and unchanged.** This is the load-bearing part: every release published before this change has no markers, and the feed is regenerated in full from every published release on every stable run. Without the fallback, the next release would blank `v0.5.1`'s description and every other historical entry.

Two guards, because both failure modes are otherwise silent: an unterminated fence would swallow the whole body, and an empty region would ship an entry telling clients nothing.

Correspondingly, **the fence always wraps the best available description** — distilled prose when there is some, the grouped PR list when there isn't. Fencing `distill` output unconditionally would put an empty region in every release cut without the key, which the guard correctly rejects, failing the release. The point of fail-open is that a missing key degrades the notes rather than blocking the release, so the composition degrades with it.

`release.yml` gets no fence: `dev-latest` is a pre-release, and `ItemsFromReleases` skips pre-releases, so it can never reach the feed.

## Verified live

Run against the real `312d532b^..4154e92c` range with a key:

- **73 commit lines → 10 grouped PR entries** deterministically (556 bytes).
- The distilled section produced five user-facing sentences (517 bytes) from 30 KB of enriched input, at roughly **$0.003** per release.
- Composed body → `ExtractNotes`: **792 bytes** extracted with a key, **566 bytes** without. No install boilerplate and no fence marker leaked into either.
- Without the key: silent, exit 0, and the workflows' new `::warning::` fires — so a key that is set but persistently failing is visible on the first release rather than looking identical to "no key configured".

Full suite passes. **The branch changes no dependency** — `git diff dev..HEAD -- go.mod go.sum` is empty.

## Requires

`GEMINI_API_KEY` as a repository secret. Absent, both workflows still produce the deterministic PR-level changelog; the prose is what goes missing.

`release-stable.yml`'s `publish` job gains `fetch-depth: 0`, without which `git describe` cannot resolve the previous stable tag and the changelog would be skipped with no error.

## Known limitations

- The distilled wording is not reproducible between runs. The grouped PR list beneath it is, so nothing is lost — but it is a change in character for a build artifact.
- Summaries occasionally include an internal-sounding item despite the prompt asking for user-visible changes only. Prompt tuning, best judged against a few real releases rather than one sample.
- `gh pr view` is one call per PR. Fine at 9; worth batching to a GraphQL query only if a dev window ever spans 50+.
- Accepted as-is: the `io.ReadAll(os.Stdin)` error branch has no automated test, and a hand-edited body with two `begin` markers would leak an HTML comment into the description.


knomit:
## Maintenance notes — traps this change leaves behind

Written for whoever (human or agent) next touches `tools/relnotes`, `tools/appcast`, or the release workflows. Each item is something that was learned the expensive way here and is not obvious from reading the code.

### Do not `go get google.golang.org/genai`

It is **already a direct dependency**, used by `internal/llm/gemini.go` — the production LLM subsystem. A bare `go get` silently bumped it v1.49.0 → v1.66.0 during this work, dragging 17 minor versions of unreviewed SDK drift into that subsystem inside an unrelated release-notes change. Reverted in f897fcc4; this branch changes no dependency at all.

If you need a newer genai, upgrade it deliberately, on its own branch, with `internal/llm` exercised.

### The fence must always wrap *something*

`ExtractNotes` rejects a fenced region that is empty — deliberately, because an empty `<description>` is a green run that tells every client nothing.

But `distill` legitimately produces **nothing** when `GEMINI_API_KEY` is unset or the call fails. So the stable workflow fences *the best available content*: distilled prose when there is some, the grouped PR list when there isn't.

**The tempting "simplification" — always fence `distill`'s output — breaks every release cut without a key.** That was caught in planning, not in code, and the shell branch that prevents it (`[ -s distilled.md ]`) is the whole point of that block.

### `ExtractNotes` must keep its unfenced fallback

A release body with no markers returns **whole and unchanged**. Every release before this change has no markers, and the feed is regenerated from *all* published releases on *every* stable run. Removing that fallback would blank `v0.5.1`'s `<description>` and every other historical entry on the next release — for shipped clients who cannot be reached to fix it.

### Match PR merges by regex, not `--merges`

`^Merge pull request #(\d+) ` — the trailing space is load-bearing. Bare `--merges` also catches branch syncs (`Merge remote-tracking branch 'origin/dev' into …`), which are merge commits but not pull requests. The reference range has 10 merge commits and 9 PRs for exactly this reason.

### `fetch-depth: 0` on `release-stable.yml`'s `publish` job is load-bearing

Without it, `git describe` cannot resolve the previous stable tag, returns empty, and the changelog is **skipped silently** via the "first release" branch — no error, no warning, just no changelog. Do not drop it while tidying checkout steps.

### `distill` is fail-open and must stay that way

`runDistill` returns `nil` on every failure path, so `main`'s `os.Exit(1)` is unreachable from that subcommand. Every failure ⇒ stderr note, nothing on stdout, exit 0. Its sibling `changes` is the opposite and must fail loudly.

It also runs under a 60s timeout: fail-open covers *errors*, not *hangs*, and the SDK sets no default deadline — an unbounded call could hold a release job to the runner's 6h cap.

### `changes` stdout must stay byte-identical

Stdout is the title-only changelog users read in the release notes. The body-enriched variant for the model goes to `-bodies-out`, from the same `Collect` result so it costs no extra `gh` calls. Do not "unify" the two renderers back into one.

### No `### Direct commits` in the reference range is correct

All 73 non-merge commits in `312d532b^..4154e92c` arrived through the 9 PR merges, so the section correctly does not render. An earlier draft of the design claimed `bump version (0abeb089)` was a direct commit — it is not, it came via PR #50. **Do not "fix" the section's absence.** Its behaviour is covered by `TestCollectSeparatesDirectCommitsFromPRCommits`.

### Free CI inference is gone (as of this writing)

GitHub Models — free inference from Actions via `permissions: models: read` and the built-in `GITHUB_TOKEN`, no secret — was the natural fit here. It was closed to new customers in June 2026 and **fully retired on 2026-07-30**: playground, catalog, inference API, and BYOK. GitHub's recommended replacement is Azure AI Foundry, which is paid, and there is no general-purpose Copilot inference API callable from CI. That is why this uses Gemini with a repo secret.

Gemini's free tier also narrowed in April 2026 — **Flash and Flash-Lite only, no Pro models**. `gemini-3.5-flash-lite` is what this uses and is verified working live.

### Costs, measured not estimated

~7,600 input tokens (30 KB of PR titles + bodies for 9 PRs) and ~200 output ⇒ **≈ $0.003 per release**. Cost is not a constraint on model choice here; upgrading to `gemini-3.6-flash` is a one-line change if summaries read poorly.

### Changelog quality now inherits PR-title hygiene

Titles flow straight into the release notes. Two live examples from the reference range:

- PR #51's title is genuinely truncated **on GitHub** with a `…` (U+2026) — the tool renders it faithfully.
- PR #46's title (`Fix/review silent envelope mismatch`) was auto-generated from its branch name because none was set.

Non-conventional titles land in `### Other`, as do conventional types outside the `feat`/`fix`/`perf`/`docs` map (`ci`, `refactor`, `chore`, …). Extending `sections` in `changes.go` is a two-line change if finer buckets are wanted.

### Accepted, deliberately not fixed

- `io.ReadAll(os.Stdin)`'s error branch has no automated test — testing it needs a seam wider than the two lines it protects.
- A hand-edited body with two `<!-- appcast:begin -->` markers before one `end` leaks the second marker's literal text into the description. It is an HTML comment, invisible when rendered.
- An `ExtractNotes` error aborts the whole feed build, not just the offending release. Fail-closed is the right direction here, and the step is documented as re-runnable.

PR #70, opened 2026-08-06
feat(web): UI improvements

knomit:
This branch started as one change to the summary panel's facet clouds and grew into a pass over the whole reading surface: the summary, the fact panel, the chrome row, the filter picker, the navigation history, and — once the UI started asking honest questions — the lens counts on the server that were answering them wrong.

35 commits, 49 files, +4588/−1147. Each commit stands alone and carries its own reasoning; this is the map.

---

## 1. The summary panel

**Facet clouds → ranked columns + a browser.** `Domains`/`Entities`/`Types` cost ~298px of a ~730px panel and pushed **Highlights** — the thing the summary exists to show — to the bottom edge with one row above the fold. They also showed 10 of 218 domains with no way to reach the other 208.

`FacetPanel` renders the three facets as ranked columns — name, hairline share bar, count — in ~187px, with eleven highlight rows above the fold. `+N more` opens a full-width **facet browser** for one facet: every value ranked, a search field narrowing it, `Esc`/Back to return. The browser is why this is a component and not a style change: the three facets do not overflow by the same order of magnitude (8 types, 218 domains, 1644 entities), so expanding in place would mean something different per column. Its list is a **fixed** height sized to clip the seventh row, so opening it doesn't shove Highlights up and searching doesn't bounce it. No API change — `internal/store/index.go` already grouped every domain and entity with no `LIMIT`.

The type census moves out of `HighlightsPanel` into `FacetPanel`, where it always belonged; `HighlightsPanel` now renders the list and nothing else. Ranking is count-desc with a **name tie-break**, so two renders of one corpus can't disagree about which of two equal-count domains leads.

**Repo rows, rebuilt.** The per-mount list was the last boxed thing in the panel — one bordered card per mount, ~330px for six, encoding nothing (a 1377-fact mount and a 9-fact mount looked identical). `RepoRows` rebuilds it from the same three parts a `FacetRow` is made of, ranks mounts by fact count with the same tie-break, and spends data the panel already fetched and threw away: `changes_7d/30d/90d` have shipped in `LensRepoStats` since the union-stats endpoint landed and nothing rendered them. Three ticks, square-rooted so a write repo out-committing a read mount tenfold doesn't flatten every other row.

Then the section was trimmed twice more — the share bar goes (six hairlines read as a second ranked block competing with the facets), and the `REPOS · 6` heading goes in favour of heading the first column `Repo`. **~330px → 188px.**

## 2. The fact panel

**233px of header became one meta line, then a pinned band.** Four stacked rows in three visual languages collapse to:

```
observation · conf 0.95 · 1 source · ● knomit-kb · ⑂ agent/… · kb/…
```

Identity → measurement → address. The mount is drawn **the dashboard's way** — a hue dot plus a plain mono name — which is what let it stay on the line instead of moving elsewhere; `hypothesis` loses its dashed box (the word, the colour and the glyph already say it three times). Structurally the type/origin chips and the Confidence/Sources boxes move **out of** `FactBody`, which is now content only: prose, tags, refs.

Then the line stopped scrolling away. `FactBand` pins the meta line and the controls **above** the title, outside the scroller — a fixed band, not `position: sticky`, which fights a padded scroll container. Pinning *adds*: the title is forced onto its own line below the values, and the one thing that yields is a path that has **already** wrapped, measured via `offsetTop` + a `ResizeObserver` on the splitter (a layout outcome, not a prop).

**Metadata becomes a property table.** Domains, entities and references were three stacked sections with a heading line each. The label is now a **column** beside its values — a key column costs no vertical space. Values flow at `auto-fill` 150px; references keep full width and one per line because a URL truncates to uselessness at a third of the width. Kind labels are gone from links and facts; `another repo` and `unresolved` stay, because those say what the text cannot. One block, one hairline rule above it, conditional on there being any metadata at all. `TagCloud` → `FacetValues` (it had stopped being a cloud and no longer had a label); the refs renderer becomes a real `RefList` component.

**Filtering is not editing.** Clicking a domain or entity did nothing on a read-only fact — which in a multi-mount lens is nearly every fact — because `TagCloud` gated the click on the same `readOnly` flag that disables retract and save. `readOnly` means "your writes do not go here"; it never made it wrong to ask which other facts share this tag. The prop goes rather than staying as a dead one that invites re-gating.

## 3. The chrome row

The filter bar rendered above the **right** pane, which reads `state.filters` and `state.freeText` exactly zero times — it governs the list on the left. It moves into the top bar, and that row becomes **controls only**: everything on it opens something.

- The two elements that merely *reported* move to `StatusFooter` — the commit (a net gain in visibility: in lens context the top bar showed it only when a fact was open *and* the view was anchored) and the lens write target, which keeps its green box on the way down.
- The mounts chip becomes the control it was pretending to be. It rendered `lens.reads.length` — the total, whatever was selected — while the real picker sat in the left panel under a `SOURCES` label. `MountsPicker` replaces `SourcesDropdown`; the left panel's block is gone. Bare at rest, an amber ratio once narrowed (`6 → 5/6`).
- The branch is trimmed to the machine: `identity.go` builds `agent/<host>-<fp8>`, whose prefix is constant and whose fingerprint only separates two agents on one host — 196px for the part carrying no information, 68px now.
- The dashboard gets a **quick search** instead: a magnifier at the end of the stats strip that expands *in place* into an input (the bar's slot above is 34px the dashboard would slide down to make, moving the numbers being read at the exact moment you reach for search). Commits on ENTER, never per keystroke. `/` opens it, `esc` closes it, both hinted in the footer.

Both contexts now land within 10px of each other — switcher, scope picker, search — so the row stops resizing when you switch context.

## 4. The filter picker, and the repo facet

**The `+` picker is rebuilt as a facet browser.** It was a 212px flyout that looked nothing like the ranked panel above it, and it misinformed: values came from `SELECT DISTINCT` with no `ORDER BY`, so opening Domain listed `agentic engineering, job state…` while `ai` — the largest domain in the corpus — sat ninth and off-screen. Now a 600px sheet: facet rail, then three fixed-height columns so it never resizes under the cursor. Ordering is **per-category** — domain/entity/path get sorted because the server gives them no order at all, while type/kind/origin keep server order, which already carries meaning. Search hides below 8 values.

**The repo facet goes** (`feat(web)!`). Scoping a lens to one mount is a *move*, not a content filter, and it already had two controls. Offering it a third time as a chip made "narrow the fan-out" look like "narrow the results", and either behaviour was wrong. First `repo:` picks were remapped to `FOCUS_LENS_SOURCE`; then the facet was removed from the picker, the autocomplete prefix set and the parse site; then the machinery it left behind (`allowRepo`, the `'repo'` chip category, the chip renderer branch, `chipRepos`, the "No sources match the repo:" empty state) was deleted.

**Chip hues.** Entities were pink in the picker and blue everywhere else. Blue was unavailable because `chipColors.type` held it for all twelve types at once, so type chips now take each type's own colour from `typeStyles` — what a Library row already does — and entity takes the vacated slot. Nothing shares a hue now.

**`policy` was invisible.** It shipped as `'#f9b4'` — a four-digit hex, which CSS reads as `#RGBA`, so it has been rendering at 27% alpha everywhere a type is drawn. It could not simply be un-truncated to `'#f9b'`: that lands 42 units from `hypothesis` on weighted RGB distance, closer than any two types have ever been. It takes `'#f97'` (hue 15 was empty). Two palette guards go with it, because both failure modes are invisible to a reader: **no accidental alpha**, and **no two types closer than 60**.

## 5. Navigation and history

- **A `NavEntry` now captures every "way of looking"** — `sort` and `lensSources`, not just path and filters. Focusing a mount used to push nothing, so Back restored the path and left the union pinned to one mount. `SET_LENS_SOURCES` still does not push, exactly as `SET_SORT` does not: a dropdown toggle refines the current view rather than moving.
- **`FOCUS_LENS_SOURCE`** collapses the sources + sort pair into one action that owns the whole intent, so history has one thing to hook.
- **A highlight now reveals.** Clicking one used to set the open fact and leave the left panel at the ontology root — you arrived somewhere with no idea where it lived. It now navigates the tree to the fact's folder, switches to the ontology view and opens the fact selected in it, in **one** dispatch, so Back undoes the whole move. Only *reveal* navigates; plain selection still moves nothing.
- **Picking a lens facet lands you on a fact.** The lens union effect never dispatched `AMEND_NAV`, so in lens context a facet click filtered the list and left the right panel on the dashboard, while the same click in repo context opened the first result. `openFirstRow` now runs on both union fetches, and deliberately **not** on `lensBrowse` — walking a tree is not searching.
- **`EXIT_SEARCH`.** The Relevance segment did nothing useful and one thing harmful: relevance is *derived*, so the dispatch couldn't change the order it claimed to set, but it did null `factPath` and overwrite `librarySort` — erasing the mode there was to go back to. It is now the way out: clears free text and content chips, keeps the path chip (location, not search), and does not touch `librarySort`, which is the whole trick.
- **The sources menu floats** over the panel instead of growing it and shoving every row out from under the cursor.
- **`rowPath()` — one rule for a row's second line.** The two lens views disagreed, and inside a directory they failed in opposite directions: the tree showed the mount alone, while Recent repeated the breadcrumb verbatim on all ten rows. Strip the mount qualifier, then strip the directory the header already names. The cut is on `dir + '/'`, never the bare prefix.
- On a 265px panel a tree leaf's mount **moves to its own line** instead of stealing the title's width, and `SourceBadge` loses the last bordered mount pill in the app.
- The ontology tree loses its per-row separators and gains a hover state — the separator was the only thing giving a row shape, so removing it alone would leave loose text. The two-line flat lists keep theirs.

## 6. Counts, paging, and scope

This is where the UI work turned into server work.

**`fix(lens)`: count with `COUNT(*)`, bound the transfer with paging.** A lens over a 1403-fact mount reported **500** while the dashboard reported 1403 for the same corpus on the same screen, and no offset reached the other 903. One conflation: the union used the size of what it *fetched* as the number of facts that *exist*. The store never conflated them — `RecentFacts` runs its own `SELECT COUNT(*)`; the handler discarded it with `entries, _, err`.

- Counts are now exact whenever exactness is computable: untruncated, the deduped union length *is* the cardinality (which is what keeps two mounts sharing a path reporting one fact). Only a truncated mount falls back to the summed per-mount `COUNT(*)` — an upper bound, off by exactly the cross-mount collisions, the same trade `/stats` already makes.
- The fixed 500 was a **window**, not a bound: a mount's 501st-newest fact was unreachable at every offset. Recency needs no window — commit timestamps are comparable across mounts — so the depth requested is `offset+limit`. `maxLensSearchCandidates` and `maxLensRecencyDepth` now name the two readings separately: a depth for relevance (ranks don't merge, RFC §7.1) and a backstop for recency.

**A chip is a filter, not a search.** Clicking `ai · 666` produced a list of 50 that would not scroll. Three faults stacked: any content chip made `searchActive` true and forced relevance (but a chip is not something to *match against*, and that route is retrieval-capped and unpaged); that left the chip in Path mode where a directory walk ignores it entirely; and `api.listLensFacts` accepted only `path/query/limit/offset/repos`, so the chip triggered a refetch and sent the identical unfiltered request — a workaround that outlived its reason by a whole handler change. All three fixed; the facet click now lands on a paged, `COUNT(*)`-backed list that scrolls.

**The header names the list and counts the server's total.** The root slot fell back to a fixed "All facts" — wrong over eight directories and zero facts under Path, and wrong under Recent where `rows.length` is how far you have scrolled (a 385-fact repo read "50", then "100"). It now says **Ontology / Facts / Matches**, and each takes its count from what it actually is: Path and Relevance from their rows (unpaged), Recent alone from the server.

**The lens dashboard follows the mounts picker.** Narrowing to "none, then agentic-engineering" showed 234 facts on the left and 2065 on the right, with columns, type counts and highlights all drawn from repos the reader had just switched off. The server always supported this — `/lenses/{lens}/stats` runs the same `narrowByRepo` — only `getLensStats` never sent the params and the effect never watched `state.lensSources`. An empty selection gets its own state rather than a fetch, since sending no repo params reads as "all".

## 7. Two server-side fixes

- **`fix(store)`: a scope with only excluded types gets its own highlights.** `kb/culture` — 3 facts, all observations — showed an empty Highlights panel. `highlightExcludedTypes` exists so the substrate doesn't bury the distilled layer (1,186 observations against 128 syntheses on core); that reasoning is about *proportion* and says nothing about a scope where the excluded types are all there is. `highlights()` re-runs without the exclusion **only** on an empty result, sharing the `ORDER BY` so a fallback list is still ranked.
- **`fix(web)`: qualify lens highlight paths.** Every highlight on a lens dashboard 404'd. `/stats` was the one lens read still emitting mount-relative paths, so a read mount's highlight asked the *write* repo for a fact it never had. One line at the merge — and **after** the dedup, never before: dedup keys on repo-relative paths because a re-rooted fork shares fact UUIDs.

## 8. Housekeeping

- **Dead code deleted**, each checked for a real consumer first: `EpisodeIcon` and its five icons, `opStyles`/`defaultOpStyle`, `ExplainIcon`/`WrenchIcon`/`TrashIcon`/`ChevronUpIcon` (orphaned by the removed `ExplainView`), `loadLastRepo`/`saveLastRepo`. `REPO_STORAGE_KEY` explicitly **stays** — `loadLastContext` still migrates from it and `saveLastContext` still writes it so a downgrade lands the user on the right repo.
- The Path picker opens at the root's **children**, not the root — the empty prefix can only ever return `kb`, so the first level cost a click to learn nothing. Sixteen folders now.
- `style(web)`: `writes →` drops off the write-target pill; the pencil glyph and green treatment already say it twice.
- `fix(web)`: the `StatusFooter` tests type-check under the real build. `npx tsc --noEmit` reads the root solution file (`"files": []` + two references) and checks *nothing* while exiting 0 — which is how a broken build shipped alongside "tsc clean". Verified with `make desktop`.

## Verification

- **876 web tests pass across 70 files** (`npx vitest run`); `tsc -b` clean; `go test ./internal/store/... ./internal/web/...` green. Lint is unchanged from baseline at 86 pre-existing problems.
- New behaviour was test-driven throughout, and the facet-panel tests were each **seen red** against a deliberately broken implementation (a no-op `+N more`, a browser that closes on the first pick, a dropped tie-break, a missing empty-facet fallback, prefix-instead-of-substring search). One case-fold test *survived* its mutation because only the query side was lowered — it now asserts both directions against real mixed-case entity names.
- Two tests that were quietly asserting the bugs fixed here were rewritten to assert the behaviour instead: `TestLensStats_HighlightsAreGlobalTopNotFirstMount` (asserted a read mount's bare path — now asserts by title) and `TestHighlights_PathPrefixScopesTheList` (asserted the empty list the exclusion was producing — now proves scoping by *which* facts come back, in both `kb/s/` and `kb/o/`).
- Nearly every commit was additionally driven live against a real corpus, with measurements recorded in its message — panel heights, pixel deltas, before/after API responses, and back-stack round trips.

**Note for reviewers:** the store and lens-handler changes are server-side, so a running desktop app serves the old binary until restarted.


knomit:
## Post-merge: knowledge-base follow-ups

Deliberately **not** done in this PR — the corpus should be updated against the merged
state, not against a branch. Listed here so it isn't lost.

### 1. New fact — a floating menu in the top bar needs `data-nodrag` on an ancestor, not on its trigger

No fact covers this and it is a trap the next top-bar control will fall into.

`TopBar` makes the whole bar a window-drag handle by posting `wails:drag` on mousedown,
bailing only when `(e.target).closest('[data-nodrag]')` matches. A menu rendered as a
**sibling** of its trigger therefore has no tagged ancestor: on the desktop build the
native drag loop swallows the click and the menu's options cannot be activated at all.
The tag has to sit on the node that contains both.

Second half of the same fact: `-webkit-app-region` is inert here. WKWebView ignores it;
Wails uses the inherited `--wails-draggable` custom property. `MountsPicker` carried a
`WebkitAppRegion: 'no-drag'` that never did anything.

Suggested shape: `kb/gotchas/ui/desktop/topbar-drag-region-ancestor`, refs
`web/src/TopBar.tsx` (`startWindowDrag`), `web/src/MountsPicker.tsx`, and the
`describe('the desktop drag region')` block in `web/src/MountsPicker.test.tsx` as
enforcing code. The switcher menu escaping via `createPortal` is the other legal way out
and belongs in the body.

### 2. `kb/invariants/web/collections/count-vs-transfer/8ee5baf2.md` — add the overflow consequence

The invariant governs `lensFanoutDepth`. This PR found a way for the bound to evaluate to
a value that is not a bound: `offset + limit` overflows to negative on an absurd offset,
which **clears** the `> maxLensRecencyDepth` test rather than tripping it and reaches the
store as `LIMIT -1` — SQLite's spelling of *no limit*. Every mount then materialises its
whole corpus, which is precisely what the constant's doc comment promises to prevent.

The consequence a consumer should act on: *a computed depth must be checked for
non-positive as well as over-cap, because a non-positive `Limit` disables the limit
instead of tightening it.* Worth naming the misreading too — "the cap is the only failure
mode" is what made a one-sided comparison look complete.

Enforcing ref to add: `TestLensFanoutDepth_OverflowingOffsetStillClamps` in
`internal/web/handlers_lenses_read_test.go`.

### 3. `kb/conventions/ui/filter-multi-chip-semantics/d6f43cd2.md` — needs an enforcing ref at the API layer

The convention (domain/type/ep OR, entity AND, path single-only) was correct and the
client quietly violated it: `api.recent` took a single `typeFilter` string, so a second
`type` chip collapsed it to `undefined` and **all** type filtering was dropped, widening
the list while both chips sat in the bar. The fact's refs are a design plan and
`web/src/state.ts` — neither of which is where the semantics are actually honoured or
broken.

Add refs for `web/src/api.ts` (`recent`, now CSV like every other multi-value facet) and
the two new cases in `web/src/Library.chrono.count.test.tsx`. The body is worth a line on
*where* OR is implemented — server-side `splitCSV` → `SearchOptions.IncludeTypes` — since
the endpoint always accepted a list and only the client was narrow.

### 4. Stale `src://` blob refs to refresh

Most of these had already drifted before this PR; it moves them further.

| Fact | File | Note |
|---|---|---|
| `kb/invariants/web/collections/count-vs-transfer/8ee5baf2.md` | `internal/web/handlers_lenses_read.go` | already drifted at HEAD; changed again here |
| `kb/incidents/lens/union-window-hid-903-facts/1d418a69.md` | `internal/web/handlers_lenses_read.go` | same blob, same drift |
| `kb/decisions/ui/summary-panel/repos-ranked-rows-with-activity/fed2c140.md` | `web/src/FacetPanel.tsx` | ref was **current**; this PR makes it stale (comment only) |
| `kb/decisions/ui/summary-panel/repos-ranked-rows-with-activity/fed2c140.md` | `web/src/RepoRows.tsx` | already drifted; tooltip changed here |
| `kb/decisions/ui/filter-bar/picker-facet-browser/1b0dfa83.md` | `web/src/FilterBar.tsx` | already drifted, untouched by this PR |

### 5. Optional — record that the facet browser closes on a pick

`FacetPanel`'s `FacetBrowser` claimed in comment that it "stays open after a pick", which
it never could: the pick adds a chip, the chip flips `path` → `recent`, the refetch
`AMEND_NAV`s row 0, and `RightPanel` draws the summary only when `!factPath`. Resolved by
correcting the comment — closing is the intended behaviour, consistent with a facet pick
everywhere else. Small, but it is the kind of thing that gets re-litigated, so it may be
worth a decision fact rather than only a code comment.

Nothing here blocks the merge.

PR #71, opened 2026-08-07
perf(store): compute a commit's diff only when the commit is actually new

knomit:
## The bug

`CommitLogSync`'s iterator returned `(hash, parents, entries)` together, so the caller had to compute the payload **before** the dedup check could reject it. In `populateCommitLog` that payload is an `object.DiffTree` of the commit against its first parent — ~307 SQLite object loads, ~2 ms per commit.

Every repo open re-walks a fully-populated DAG, so nearly every commit is a dedup hit. The cheap check was gating the expensive work, in the wrong order:

```
store.Open        =   4.0 ms
OpenRepo          = 4.187 s      <- 1831 diffs computed, all discarded
  iter.Next       =  12.5 ms  ( 0.3%)
  changedFiles    = 4.065 s   (99.5%)   <- object.DiffTree
  commitEntries   =  0.45 ms  ( 0.0%)
  dedup SELECT    =   8.9 ms  ( 0.2%)
```

## The fix

Make the payload a thunk:

```go
type CommitLogPayload func() (parents []string, entries []CommitLogEntry, err error)

func (s *Storer) CommitLogSync(branchName string,
    iter func() (hash string, payload CommitLogPayload, err error)) error
```

`CommitLogSync` calls `payload()` only after the `branch_commits` lookup misses. `AppendCommitLog` gets the same shape and now also skips reading the commit object when the commit is already recorded.

**The DAG invariant is untouched.** A dedup hit still skips-and-continues, never short-circuits — every commit is still visited and merge-parent lines are still followed. Only the per-commit *payload* became lazy. `TestCommitLogSync_WalksPastDedupHits` anchors this.

## Results

| | before | after |
|---|---:|---:|
| `OpenRepo` on core.db (1831 commits) | 4.187 s | **68.6 ms** |
| Boot, 24 repos (run 1) | 35.42 s | **3.02 s** |
| Boot, 24 repos (run 2) | 37.86 s | **1.85 s** |

Both binaries built from this same tree with the fix stashed for the "before" build. The baseline reproduces the previously-measured 35.34 s.

## Correctness

Verified identical output, not just faster. After booting all 24 repos on each binary, `branch_commits` / `commit_log` / `commit_parents` row counts and an md5 over every `(commit_hash, path, action)` row are **byte-identical** across all four repo shapes:

```
core-1                 5482 bc, 4904 cl, 1856 cp, md5=106b2bc7...  (both)
knomit-kb-1            1607 bc, 1652 cl,  783 cp, md5=0243f843...  (both)
agentic-engineering-1   552 bc,  523 cl,  187 cp, md5=6733ded7...  (both)
knomit-io-kb-1           20 bc,   12 cl,    8 cp, md5=e785b982...  (both)
```

Full `go test ./...` green. The new tests are non-vacuous — re-run with the eager ordering restored, both fail with `changedFilesInCommit: tree: object not found`. The end-to-end anchor deletes the tree objects of recorded commits, so any attempt to diff them fails outright; a re-walk succeeding proves the diff never happens.

## Two corrections to the earlier diagnosis

- **Not zlib, and not per-round-trip latency.** Objects are stored uncompressed (`storer.go` inserts the raw blob). Individual loads are 6.9 µs — SQLite is fine. The cost is the *number* of object loads.
- **Not `commits × branches`.** `OpenRepo` walks only the HEAD branch, and `healIndexBranches`' `SyncLocked` does not populate the commit log (only the `stale` path's `Rebuild` does). Cost is linear in **distinct commits on HEAD**. A repo used from five machines does not pay 5×.

### DIFF
--- a/internal/store/branch_commit_log.go
+++ b/internal/store/branch_commit_log.go
@@ -34,27 +34,35 @@ func (rh *repoHandler) populateCommitLog(ctx context.Context, branch string) err
 	}
 	defer logIter.Close()
 
-	var count int
-	err = rh.gits.CommitLogSync(branch, func() (string, []string, []storegit.CommitLogEntry, error) {
+	// The payload is a thunk, not a value: changedFilesInCommit is an
+	// object.DiffTree costing ~300 object loads (~2 ms) per commit, and on a
+	// warm open every commit here is already recorded. CommitLogSync calls the
+	// thunk only for commits it will actually insert, so a no-op re-walk of a
+	// populated branch costs one indexed lookup per commit instead of a diff.
+	var count, computed int
+	err = rh.gits.CommitLogSync(branch, func() (string, storegit.CommitLogPayload, error) {
 		c, err := logIter.Next()
 		if err == io.EOF {
-			return "", nil, nil, nil
+			return "", nil, nil
 		}
 		if err != nil {
-			return "", nil, nil, err
+			return "", nil, err
 		}
 		count++
-		files, err := changedFilesInCommit(c)
-		if err != nil {
-			return "", nil, nil, err
-		}
-		return c.Hash.String(), parentHashes(c), commitEntries(c, files), nil
+		return c.Hash.String(), func() ([]string, []storegit.CommitLogEntry, error) {
+			computed++
+			files, err := changedFilesInCommit(c)
+			if err != nil {
+				return nil, nil, err
+			}
+			return parentHashes(c), commitEntries(c, files), nil
+		}, nil
 	})
 	if err != nil {
 		return fmt.Errorf("populateCommitLog: sync: %w", err)
 	}
 
-	log.Debug().Int("commits", count).Msg("commit_log: populated")
+	log.Debug().Int("commits", count).Int("computed", computed).Msg("commit_log: populated")
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


knomit:
## Post-merge actions

### 1. Chase the ~307 object loads per commit — the cold paths still pay it

This PR removes the diff from the **warm** path only. The diff itself is untouched and still costs ~2 ms/commit wherever it genuinely runs:

- **First populate** of a newly cloned/created repo (one-time, but it is the first-run impression)
- **`Rebuild` phase 0** via `rebuildCommitLog` — measured at 4.9 s of a 14.9 s rebuild, the largest component when embeddings are intact
- **Migrations** that clear `commit_log` and force a re-walk

Two independent leads, both measured here but not yet chased:

- **`merkletrie` is not pruning.** It is supposed to skip subtrees whose hashes match, so a one-file commit should cost ~10 tree loads, not ~307. Worth confirming against `plumbing/object/treenoder.go` — note `transformChildren` builds a noder for *every* entry of a tree, which may be defeating the prune for wide trees.
- **`Storer` has no object cache at all.** Consecutive commits diff against trees they just loaded — commit N−1's tree *is* commit N's parent tree. A small LRU in front of `EncodedObject` would help every git path, not just this one. go-git's filesystem storer uses `cache.ObjectLRU` for exactly this; ours does not.

### 2. Update the scratch note

`.claude/harness/scratch/2026-08-07-repo-open-slow-commitlog-walk.md` carries two claims this PR disproves. Either correct it in place or mark it superseded:

- "per-commit object load + zlib decode, one round trip each" — objects are stored uncompressed; there is no decode, and per-load latency (6.9 µs) is not the problem
- "open ≈ 1.2 ms × commits × branches" — cost is linear in **distinct commits on HEAD**; `OpenRepo` walks one branch

Its first open question ("does open cost track distinct commits or `branch_commits` rows?") is now answered: distinct commits on HEAD. Its second ("is `healIndexBranches` contributing?") is answered too: no — `SyncLocked` does not populate the commit log, and `OpenRepo` accounts for ~4.19 s of the 4.19 s. Its third (does `Rebuild` phase 0 share the hot loop?) is **yes**, and is item 1 above.

### 3. Record the finding in knomit

Worth a `/knomit-remember` once item 1 is settled, so one fact captures the whole shape rather than two partial ones. The durable lesson is broader than this call site:

> A cheap dedup check cannot save you if the expensive work is computed before the check runs. When an iterator contract returns a payload alongside its key, the payload is unconditionally paid.

That is a reusable anti-pattern, and `CommitLogSync` is unlikely to be the only place the shape appears. Also worth linking from the existing `commitlog-sync-walks-dag-not-shortcircuit` invariant: that fact correctly forbids ending the walk early, and this PR is the worked example of the distinction it implies but does not spell out — **you may not make the walk shorter, but you may make each step cheaper**.

### 4. Re-baseline the lazy-load / idle-unload tradeoff

The 4.4 s stall on first touch of a cold tenant was the main argument against idle-unload. That stall is now ~70 ms. Unloading `core` still only reclaims ~19 MB of a ~900 MB RSS (the memory is the ONNX model, not the repo), so the *memory* side of the trade is unchanged — but the *latency* objection is gone and the decision is worth revisiting on its merits.

PR #76, opened 2026-08-09
fix(bridge): emit hookEventName so CC stops discarding our nudges

knomit:
## Problem

Every edit was producing this in the transcript:

```
PostToolUse:Edit hook error
Hook JSON output validation failed — hookSpecificOutput is missing required field "hookEventName"
```

`emitAdditionalContext` built a `hookSpecificOutput` object containing only
`additionalContext`. Claude Code requires `hookEventName` inside that object and
**discards the entire payload** when it is absent — so the nudge never reached
the model. The hook exited 0 and produced output, which is why this went
unnoticed: it looked like it was working.

Confirmed against the installed Claude Code binary (2.1.226) rather than from
memory — its embedded hook docs state `hookSpecificOutput` "must include
`hookEventName`", and the exact error string lives in its validator.

## Scope

All three JSON-emitting hooks shared the broken helper:

| Hook | Wired to | `hookEventName` |
|---|---|---|
| `post-edit` | `PostToolUse` (matcher `Edit\|Write\|MultiEdit`) | `PostToolUse` |
| `post-ask` | `PostToolUse` (matcher `AskUserQuestion`) | `PostToolUse` |
| `pre-compact` | `PreCompact` | `PreCompact` |

`session-start` was unaffected — it writes plain text to stdout rather than JSON.

## Fix

`emitAdditionalContext(w, event, ctx)` now takes the event name, and each caller
passes the CC event it is wired to in `settings.json`. The doc comment calls out
the trap: it is the **settings.json event name**, not the bridge's own
subcommand name (`post-edit`, `pre-compact`, …), which is the easy way to get
this wrong and land right back at a silently-dropped payload.

## Verification

- Wrote the three assertions **first** and watched them fail with
  `hookEventName = "" want "PostToolUse"` / `"PreCompact"` — the bug reproduced
  exactly before any production code changed.
- `go test ./tools/bridge/claude/...` passes; `go vet ./tools/bridge/...` clean.
- Ran the rebuilt binary end-to-end against a real post-ask payload; it now
  emits `{"hookSpecificOutput":{"hookEventName":"PostToolUse","additionalContext":"…"}}`.

## Notes

- **Not live until `dist/knomit-bridge` is rebuilt** — `~/.local/bin/knomit-bridge`
  symlinks to it, so the error persists locally until then.
- `go build ./...` fails in a fresh worktree on `tools/calibrate` and the root
  package with `library 'tokenizers' not found`. That is the gitignored
  `dist/darwin-arm64/lib` being absent from a new worktree, pre-existing and
  unrelated to this change.

### DIFF
--- a/tools/bridge/claude/helpers.go
+++ b/tools/bridge/claude/helpers.go
@@ -211,10 +211,17 @@ func agentBranch(repo string) string {
 // Returns nil if ctx is empty (caller can short-circuit before any output).
 //
 // event MUST be the CC hook event this hook is wired to in settings.json
-// ("PostToolUse", "PreCompact", …) — NOT the bridge's own subcommand name.
-// CC validates hookSpecificOutput and drops the entire payload with "Hook JSON
+// ("PostToolUse", "Stop", …) — NOT the bridge's own subcommand name. CC
+// validates hookSpecificOutput and drops the entire payload with "Hook JSON
 // output validation failed" if hookEventName is missing or does not match, so
 // getting it wrong silently costs the nudge, not just the field.
+//
+// Naming the wired event is necessary but NOT sufficient: hookSpecificOutput is
+// a discriminated union, and only these events carry an additionalContext
+// variant — PreToolUse, UserPromptSubmit, PostToolUse, PostToolBatch, Stop,
+// SubagentStop. Every other event (PreCompact, SessionStart, SessionEnd, …)
+// fails the same validation no matter what is passed here; those hooks must
+// emit plain text on stdout instead — see hookPreCompact and hookSessionStart.
 func emitAdditionalContext(w io.Writer, event, ctx string) error {
 	if ctx == "" {
 		return nil
--- a/tools/bridge/claude/helpers.go
+++ b/tools/bridge/claude/helpers.go
@@ -209,15 +209,23 @@ func agentBranch(repo string) string {
 // emitAdditionalContext writes a JSON object to w that injects ctx as a
 // system reminder via CC's hookSpecificOutput.additionalContext mechanism.
 // Returns nil if ctx is empty (caller can short-circuit before any output).
-func emitAdditionalContext(w io.Writer, ctx string) error {
+//
+// event MUST be the CC hook event this hook is wired to in settings.json
+// ("PostToolUse", "PreCompact", …) — NOT the bridge's own subcommand name.
+// CC validates hookSpecificOutput and drops the entire payload with "Hook JSON
+// output validation failed" if hookEventName is missing or does not match, so
+// getting it wrong silently costs the nudge, not just the field.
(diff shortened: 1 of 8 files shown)


knomit:
## Post-merge actions

Review of this branch found that the `PreCompact` third of the original fix was
still broken; `e2b22aa0` closes it (plain stdout instead of a `hookSpecificOutput`
envelope, a test that asserts the real contract, and a corrected doc comment on
`emitAdditionalContext`). `dev` is merged in and the branch is green.

Three things still need doing after this lands:

**1. Rebuild the installed bridge — the fix is inert until you do.**
The hooks in `.claude/settings.json` shell out to `knomit-bridge` on `PATH`, which
is a symlink to `dist/knomit-bridge` — a binary built from whatever was checked out
at the time, not from this branch. Until `make build` runs on the merged `dev`,
every hook here keeps running the old code and all three nudges stay dead. Anyone
else with the bridge installed needs the same rebuild.

**2. Confirm `PreCompact` on a real compaction.**
Everything in this PR is verified except that CC *accepts* the new payload — that
needs an actual compaction, which I could not trigger. The evidence so far is that
the failure mode is gone (no JSON left to reject) and the output shape matches what
`executePreCompactHooks` reads as `newCustomInstructions`. On the next compaction
after the rebuild, check for two things: no `PreCompact [knomit-bridge claude hook
pre-compact] failed: Hook JSON output validation failed …` line, and the
capture-candidate list showing up in the post-compaction instructions. If the nudge
is silently absent with no error, the transport is fine and the problem is upstream
in `matchIntents` / `scanTranscript`, not in the envelope.

**3. Record the schema constraint in knomit.**
`hookSpecificOutput` is a discriminated union, and only six events carry an
`additionalContext` variant: `PreToolUse`, `UserPromptSubmit`, `PostToolUse`,
`PostToolBatch`, `Stop`, `SubagentStop`. Every other event — `PreCompact`,
`SessionStart`, `SessionEnd` — fails validation regardless of what `hookEventName`
says, and CC's failure mode is to discard the hook's entire result, so the cost is
the whole nudge rather than one dropped field. This bit us twice: once as the
original bug, once as the incomplete fix. It is written down in the
`emitAdditionalContext` doc comment now, but that only helps someone already editing
that function — it belongs in the corpus as a gotcha under `gotchas/integrations/`
so it surfaces on recall before the next bridge hook gets written.

### Note on the test that hid this

The superseded `TestHookPreCompact_EmitsHookEventName` round-tripped our own JSON
encoder and asserted `hookEventName == "PreCompact"` — an invariant CC does not
honour. It passed green while the runtime payload was rejected outright. A test
that only checks that we wrote what we intended to write cannot detect that the
other side rejects it; the assertion has to name the consumer's contract. The
replacement asserts the payload is *not* JSON and carries no `hookSpecificOutput`
key, which is a claim about what CC will accept rather than about our own encoder.

PR #77, opened 2026-08-09
fix(web): move the description pencil into the block heading

knomit:
## The bug

The edit affordance for a repo's **Description** (and a lens's **Note**) sat beside the prose as a flex sibling. The rendered body is a fixed-height scroller (`maxHeight: 360`), so that sibling reserved its column for all 360px of it — a README long enough to scroll wrapped early down its *whole* length to leave a gutter for a 13px glyph that is only ever visible at the top.

## The fix

The pencil moves into the block heading's right-aligned action slot — where **Read mounts** already puts its own pencil — and the prose gets the full column width.

Supporting changes:

- `editing` lifts to the caller (`RepoDetail`, `LensDetail`) so the heading and the body stay one control. `DescriptionEditButton` is the heading half.
- The draft becomes `string | null`, falling through to `markdown` when untouched. This removes the seeding step the split would otherwise have needed, keeps Cancel discarding by construction, and avoids adding two new `react-hooks/set-state-in-effect` errors — lint stays at the pre-existing baseline of 2 in this file.
- No reset plumbing: both detail panes are already keyed by name, so they remount on switch.
- The pencil is *disabled* rather than unmounted while the editor is open, so the heading does not reflow mid-edit — same treatment as the Read mounts pencil.

## Verification

- `npx vitest run` — **940 passed** (938 before, + 2 new).
- The two new tests pin the placement and the disabled-while-editing state. Both were confirmed to **fail** against the old layout by running them on `dev`'s `RepoManager.tsx`.
- `npx tsc -b` clean; `npx eslint` at baseline.
- Confirmed live in Chrome against a running server: prose spans the full column, pencil sits on the `Description · README.md…` line, opens/dims/restores correctly.

### Not verified visually

The lens **Note** block was not seen in a browser. The local server was built from `feat/repo-registry-control-db`, which returns `LensRead.repo` as `{uid, name}`; on `dev` that type is still `string`, so any `dev`-based frontend crashes on the lens page. Pre-existing version skew, unrelated to this change — the Note block uses the same component and its unit tests pass.

### DIFF
--- a/web/src/RemoteStatus.tsx
+++ b/web/src/RemoteStatus.tsx
@@ -80,12 +80,12 @@ export function RemoteCard({ repo, agentBranch, readOnly, state, onConnect, onDi
           <div style={{ display: 'flex', alignItems: 'center', gap: 2 }}>
             <button type="button" className="k-bare" data-testid="remote-reconnect"
               title="Reconnect / change remote" aria-label="Reconnect or change remote"
-              style={cardIconBtn} onClick={onConnect}>
+              style={cardIconBtn()} onClick={onConnect}>
               <PencilIcon color="#888" size={13} />
             </button>
             <button type="button" className="k-bare" data-testid="remote-disconnect"
               title="Disconnect remote" aria-label="Disconnect remote"
-              style={cardIconBtn} onClick={onDisconnect}>
+              style={cardIconBtn()} onClick={onDisconnect}>
               <UnlinkIcon color="#a66" size={13} />
             </button>
           </div>
--- a/web/src/RepoManager.test.tsx
+++ b/web/src/RepoManager.test.tsx
@@ -1,7 +1,7 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { render, screen, waitFor, fireEvent, within } from '@testing-library/react';
 import { RepoManager } from './RepoManager';
-import { api, MAX_LENS_DESCRIPTION_BYTES } from './api';
+import { api, MAX_LENS_DESCRIPTION_BYTES, MAX_REPO_DESCRIPTION_BYTES } from './api';
 
 // Only `api` is stubbed. The module's other exports — the description byte
 // caps — pass through from the real module, so a test asserting on a cap is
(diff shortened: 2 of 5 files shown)


knomit:
## Post-merge activities

**1. Record the placement rule in knomit.** A `/knomit-decided` fact for the manage-page blocks: *a block-level affordance belongs in the heading's action slot, not as a flex sibling of the body* — with the reason that a fixed-height scrolling body turns a sibling into a full-height reserved column. Read mounts already followed this; Description was the outlier, and nothing in the corpus said so. Recall on `manage page block layout pencil edit affordance` returned nothing, which is why this had to be re-derived.

**2. Repair a stale fact.** `kb/decisions/repos/description-from-kbmd/8992c2a9.md` still lists `DescriptionCard` and `Disclosure` as entities and describes the description as "rendered collapsible in the detail pane". The disclosure was removed before this PR and the card is now a `SettingsPage` section. Candidate for `/knomit-update` (body + entities + a fresh `src://` ref), not retraction — the README-at-HEAD claim it exists to make is still true.

**3. Verify the lens Note block visually.** Not done here: the local server was built from `feat/repo-registry-control-db`, which returns `LensRead.repo` as `{uid, name}`, while `dev` still types it `string` — so a `dev`-based frontend crashes on the lens page before the Note block renders. Worth one look once frontend and server are on the same branch. Same component, same tests, so this is confirmation rather than suspicion.

**4. Sweep for the same shape elsewhere.** This bug is "icon button as flex sibling of a fixed-height scroller". Worth grepping the manage surfaces for other `display: flex` rows whose long child has a `maxHeight` + `overflowY`, since each one silently narrows its content for its whole height.

**5. Pre-existing flake, unrelated to this PR.** `src/Library.keyboard.test.tsx` — *"dispatches SET_FACT_PATH on ArrowRight"* — fails roughly 1 run in 6 of the full suite, and reproduces on unmodified `dev` at the same rate (1/6 there, 1/6 with this branch). Passes every time in isolation, so it is load/scheduling sensitive rather than order-dependent on a specific file. Deserves its own issue.

**6. Housekeeping.** Delete the `fix/description-edit-placement` worktree and branch after merge.

knomit:
### Post-merge actions

**One, and it lands on #75.**

This PR turns `cardIconBtn` from a `CSSProperties` object into `cardIconBtn(disabled = false)` (`web/src/manageStyles.ts`), matching the shape `btn(disabled)` already had. Inline styles can't express `:disabled`, so the cursor has to be switched by hand.

`feat/repo-registry-control-db` (#75) is currently behind `dev` and carries three bare call sites:

- `web/src/RemoteStatus.tsx:83`, `:88` — merge clean (that file isn't touched by #75), no action
- `web/src/RepoManager.tsx:831` — the description pencil in its **old in-body position**, which this PR moves into the block heading; expect a real conflict here, not a mechanical one
- `web/src/RepoManager.tsx:1149` — the lens read-mounts pencil, whose disabled expression is hoisted to `readsLocked` here

So when #75 merges `dev`: resolve the description-block conflict toward the heading placement, then make every surviving `cardIconBtn` reference a call — `cardIconBtn()` where the button is never disabled, `cardIconBtn(<expr>)` where it is. `tsc -b` catches any that are missed, so this can't ship silently.

**Nothing else.** The change is web-only: no migrations, no config or API surface, no docs referencing pencil placement.

PR #82, opened 2026-08-10
feat(web): connect remote becomes a sub-page of the repo, not a surface takeover

knomit:
Re-targeted from #80, which was opened against `master` by mistake. Same four commits, unchanged; this repo lands work via `dev` → `master`.

Connect-remote moves from a full-dialog takeover to a sub-page of the repo, with the exit gated while a commit is in flight (a commit is a store swap and must not be abandoned mid-write).

Includes the fixes from review: `created`-flag stale-closure fix in `handleTest`, `doneTimerRef` cancellation on unmount, the `commitFailed`/`leavable` split, and a `leftRef` reset on effect setup so StrictMode cannot pin it true.

Known follow-ups, deliberately not in this PR:
- A commit stream that ends without a `done`/`error` event leaves the reader unable to exit Manage (no timeout, no escape hatch).
- Leaving the sub-page via the repo rail does not delete the server-side session, so a temp clone leaks until the 10-minute idle sweep.

### DIFF
--- a/web/src/App.manage.test.tsx
+++ b/web/src/App.manage.test.tsx
@@ -41,6 +41,14 @@ vi.mock('./api', async (importOriginal) => {
   return {
     ...actual,
     fetchVersion: vi.fn().mockResolvedValue({ version: '0.0.0', commit: 'abc', full: '0.0.0.abc', readOnly: false }),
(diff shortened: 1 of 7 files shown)


knomit:
Superseded by #83, which brought the same content into dev via master.

PR #86, opened 2026-08-11
feat: consolidate knomit-owned private data under .knomit/, with agent-writable state

knomit:
Consolidates knomit-owned private data under a single `.knomit/` root and makes subdirectories under it writable through the fact tools, so periodic jobs can keep state without polluting the fact corpus.

## The problem

The `agentic-engineering` crawl job stored its bookkeeping as real facts under `kb/meta/`. Because they were real facts they were indexed, embedded, ranked in `knomit_query`, browsed in the UI, exported by OKF and swept by synthesis. Job state is machinery, not knowledge.

Moving it to a dot-prefixed path did not work, because `fact.IsPrivatePath` was one predicate doing two jobs: excluding paths from discovery at 7 walkers, *and* refusing them at 5 write sites. The refusal is right for a stray draft and wrong for deliberate job state, which *wants* to be invisible to readers.

## The change

Discovery exclusion is **unchanged** at all 7 walkers. A new `fact.IsWritablePrivatePath` governs writes:

> Agent-writable = `.knomit/<area>/…`, at least one subdirectory deep, minus a reserved list.

An **allow-list rooted in ownership**, deliberately not a deny-list of foreign dot-roots. A deny-list fails open: the moment a repo grows `.vscode/` or `.devcontainer/`, knomit would be authorized to write another tool's territory. Nothing in the code knows the word `jobs` — `<area>` is any name a caller picks, so a new category of private data needs no release.

- `knomit_learn` gains an optional `path` for named slots, mutually exclusive with `topic`/`category`, failing if the slot already exists.
- `knomit_update` and `PUT /facts/{path}` relax to match.
- `knomit_retract` and `DELETE /facts/{path}` **gain a guard they never had** — a delete is a write, and both previously accepted any private path that create and update refused. Intentional behaviour change.
- The ontology moves to `.knomit/ontology.yaml`; `.domains/ontology.yaml` is kept as a legacy read and the pre-dot `domains/` path is dropped. `okf/source`'s duplicated copies of those literals are folded onto shared constants in `fact`.

## Migration

**No migration code, by design** — see the caveat in the post-merge comment below, which corrects the naive `git mv` procedure. A repo that is not migrated keeps loading its own ontology from the legacy path and logs a WARN; it does **not** silently fall back to the embedded default, which would validate new facts against the wrong taxonomy.

## Testing

All 7 discovery guards are pinned by tests that are individually falsifiable — each fails if its guard line is deleted. That took a second round: `.knomit/` sits at the repo root, so the ontology-root prefix check rejects it before `IsPrivatePath` is consulted, meaning `.knomit/`-only assertions would have survived deleting the very guards they existed to protect. `kb/.drafts/` companion fixtures isolate the guard.

`go build`, `go vet` and `go test ./...` are clean.


knomit:
## Post-merge actions

Deferred deliberately until the code was reviewed, so a design that changed under review would not leave stale records behind.

### 1. Decisions to capture (`/knomit-decided`)

1. Job state is addressed by **named slots at caller-chosen paths**, not UUID leaves found by prefix query. Rejected: UUID leaves plus a private-visible `knomit_query` mode; named slots plus a list tool.
2. All knomit-owned private data consolidates under **`.knomit/`**, with the carve-out that anything a git provider resolves by exact name (`README.md`, `LICENSE`, `.github/`) stays at the tree root.
3. Write authorization is an **allow-list rooted in ownership**, chosen over a deny-list of foreign dot-roots — a deny-list fails open as a repo grows `.vscode/`, `.devcontainer/` and friends.
4. State files carry the **full fact envelope**, and `knomit_learn` gains an optional `path`. Rejected: a lenient `ParseFact` mode; `knomit_update` as an upsert.

### 2. Corpus facts this change makes false

- `kb/invariants/fact/private-paths/b6babb00.md` — says "refused at five write sites". Now **seven**, and all of them carry the `.knomit/<area>/` exception: `learn`, `update`, `PUT facts`, plus `retract` and `DELETE facts` which previously had no guard at all. `handlers_fact_create.go` and `synthesize/decision.go` are unchanged.
- `kb/decisions/fact/kb-repo-layout/private-dot-paths/b3425b43.md` — states the rejection-on-write rationale unconditionally; it now holds only outside `.knomit/<area>/`. Also names `.domains/` as the ontology location.
- `kb/decisions/repos/kb-repo-layout/ontology-dot-domains-fallback/318f619a.md` — both path constants changed and the pre-dot `domains/` fallback is dropped.

These were deliberately **not** updated from the branch: their convention is post-merge verification against merged blob refs.

### 3. Per-repo ontology migration — read this before trying `git mv`

**The obvious procedure does not work.** knomit stores each repo's git objects inside SQLite (`~/.knomit/repos/*.db`, via `storegit.NewStorer`); there is no filesystem worktree to `git mv` in. Nor can the REST API do it — `PUT .../facts/.knomit/ontology.yaml` is refused by the depth rule, correctly, since the ontology is server-owned. There is no CLI command for it either.

Real options:

- **Do nothing.** A repo on `.domains/ontology.yaml` keeps loading *its own* ontology and logs a WARN naming the canonical path. This is a supported steady state, not a broken one.
- **Round-trip through the remote**, for repos that have one: clone, `git mv .domains/ontology.yaml .knomit/ontology.yaml`, push, let knomit sync.

Whichever is chosen, note the behaviour that makes the WARN safe: `loadOntology` remembers which path it read, and the boot-time preset refresh writes back to *that* path — so a legacy repo never grows a second ontology file with the stale one indistinguishable from the live one. That is pinned by `TestLoadOntology_PresetRefreshWritesBackToThePathItRead`.

### 4. `agentic-engineering` job cutover — before the next crawl run

- Copy the `crawl-sources` HEAD body to `.knomit/jobs/agentic-engineering/crawl-sources.md`.
- Walk `crawl-state`'s history, union every revision's URL list, and write that union as revision 1 of `.knomit/jobs/agentic-engineering/crawl-state.md`. **Do not attempt a rename** — `RevisionsBefore` keys on an exact path and does not follow renames, so an orphaned history means a silent full re-crawl with duplicate facts across 30+ URLs.
- `knomit_retract` the three old `kb/meta/jobs/agentic-engineering/**` facts.
- Drop the `spec/` fact mirror rather than moving it: the spec is attached to every prompt as Appendix S, its "self-describing clone" purpose stops working under `.knomit/`, and moving it into `.knomit/jobs/` would make it structurally writable while the spec says never write it.
- Apply the `spec.md` / `crawl.md` / `init.md` rewrites: the state-lookup step and the "expect exactly one fact per directory" hazard both disappear with named slots. Those files are untracked (`.gitignore` ignores `.claude/*`), so they are not part of this PR.

knomit:
## Second review round — one HIGH closed

A second review pass after the PR opened found five issues. Two were holes in the **design**, not slips in the implementation. All five are fixed in `eb35bbe6..7001a8f0`; `go build`, `go vet` and `go test ./...` remain clean.

### HIGH — an agent could destroy the ontology

`IsWritablePrivatePath(".knomit/ontology.yaml/x.md")` returned **true**. The depth rule protected the ontology as a *filename* but not against that name being reused as a *directory*, and `buildTree`/`upsertEntry` replaces a same-named blob with a tree regardless of mode. So a single `knomit_learn` orphaned `.knomit/ontology.yaml`, and the next boot fell through to `fact.DefaultOntology()` — every subsequent fact validated against the wrong taxonomy. Reachable through `PUT …/facts/…` too.

That directly falsified the guarantee this PR's own comments made ("no agent can rewrite it through the fact tools").

**Fix:** the `<area>` segment may now contain no `.` at all. That covers every current *and future* server-owned loose file with no list to maintain — the same structural approach the depth rule already uses. Only the area segment is checked, so `.knomit/runs/2026.08/x.md` stays writable. Pinned end-to-end in both the MCP and web layers.

### MEDIUM — the ontology fallback chain had lost its oldest rung

`.domains/` was introduced on **2026-08-05**, six days before this branch. Dropping the pre-dot `domains/ontology.yaml` rung assumed `.domains/` was long-established; it was not. Any repo created before that date and not hand-migrated *twice in six days* would have silently loaded the embedded default — the exact failure the fallback exists to prevent.

**Fix:** three rungs, ordered canonical → `.domains/` → `domains/`, centralised in `fact.OntologyPathsNewestFirst()` so `repos` and `okf/source` cannot drift. `loadOntology`'s `srcPath` write-back is preserved and now pinned for the third rung too.

### Also fixed

- Two facts sharing an explicit `path` in one `knomit_learn` call silently collapsed into one while the response reported both as written. New with explicit paths — UUID leaves could never collide.
- `IsWritablePrivatePath` authorised paths containing `..` that `store.validatePath` then rejected at write, taking the all-or-nothing batch down. The predicate now agrees with the storage layer.
- `README.md`, `LICENSE` and the ontology rungs are now refused by the PUT and DELETE fact endpoints. Worth noting the reviewer's finding here was **partly wrong and the implementer corrected it with evidence**: `DELETE …/facts/README.md` never removed the manifest (`deleteFile` lowercases, go-git's lookup is exact). The real exposures were a planted lowercase `readme.md` sibling — which GitHub renders, its lookup being case-insensitive — and `domains/ontology.yaml`, restored above, which had no protection at all.

### Correction to the post-merge list

`kb/invariants/fact/private-paths/b6babb00.md` and the three related decision records are **all** stale now, not just the ones named earlier — the write-refusal count, the `.domains/` location, and the fallback-chain description have each changed again.

PR #89, opened 2026-08-13
Deprecate migrate-registry; make verify report the truth about a real home

knomit:
Two changes, one theme: the tooling should tell you the truth about your home.

## 1. `migrate-registry` deprecated

It converts a home shape predating the control.db repo registry. A home built by a current build never needs it; a converted one never needs it again.

The deprecation is announced on both surfaces an operator can meet: `Short`/`Long` for whoever reads `--help`, and a banner printed before the plan on every run (including `--dry-run`) for whoever pastes the command out of an old runbook. It says what waiting costs — once the command is gone, the only path off the legacy shape is re-cloning each knowledge base from its remote, and a repo with no recorded remote cannot be recovered that way.

Deliberately **not** cobra's `Deprecated` field: that makes `IsAvailableCommand()` false and drops the command from `knomit --help`. This is the command a legacy home's boot refusal names, so hiding it would leave that refusal pointing at something unfindable. A test pins that it stays listed.

## 2. Verify

On a copy of a live five-repo home, `knomit verify --all` reported **8128 integrity errors and exited 1. All 8128 were false** — and zero issues were on any branch the instance actually indexes.

Verify demanded SQLite parity for every `refs/heads/*` ref. The indexer maintains a strictly smaller set (`setupIndex`: agent branch + `upstreamMain`). Everything else produced a wall of errors: 6762 on `okf/*` refs left by the server-side export removed in bf9becbe, 1366 on another machine's agent branch arriving by fetch. bf9becbe removed the *producer* and pinned the invariant for new repos, but shipped no cleanup — so on any home older than 2026-07-25 the Critical it fixed was still live. The only two repos that verified clean were the two created after it.

| | before | after |
|---|---|---|
| issues across 5 repos | 8128 (all false) | 12 warnings, 0 errors |
| exit code | 1 | 0 |
| `--all` wall clock | 13s | 3.9s |

### Scope fix
Parity now runs only over maintained branches. The oracle is the union of `meta.last_commit:<branch>`, HEAD's target, and `meta.agent_branch_owner` — **not** `graph_schema_version:<branch>` (migration 000016 backfilled it onto never-indexed branches) and **not** the `branches` table (a fetch populates it for branches nothing indexes). HEAD is in the union because a fresh repo has no `meta` rows at all, so `last_commit` alone would make a new repo check nothing and report CLEAN — the same defect inverted. `--all-branches` opts back in.

Unindexed refs are named in the report header rather than warned about per branch; a permanent warning on every healthy repo would be the same cry-wolf failure in new clothes. Only generated `okf/*` refs warn, because those have a repair: `--prune-generated-refs` removes them and their `okf:marker:*` rows and nothing else. Verify itself stays read-only.

### New checks
`commit-parents`, `commit-log` orphans (the `commit-log` category had stopped reading `commit_log` entirely), `derived-tables`, `PRAGMA quick_check`/`foreign_key_check`, `schema-version`, `orphan-objects`, `embedding-identity`.

Two calibrated from measurement rather than intuition:
- `fact_domains` ↔ `fact_domain_tokens` compared at `fact_id` granularity, not `(fact_id, domain)`. Both writers insert `canonicalizeDomain(d)` so the strings "must" match — on a live repo they don't (`claude-code` vs `claude code`), because the function gained a de-hyphenize step and only the token side was repopulated. The string-level join reports 6 phantom failures on a healthy repo.
- `orphan-objects` is a warning and a count. Unreachable objects accumulate normally.

### CLI
Usage errors exit 2 (were 1, colliding with "issues found"); `cmd.Context()` so SIGTERM can actually cancel; app lifetime moved into a helper so `os.Exit` stops skipping `defer a.Close()`; `--json`; `--max-issues`; deterministic ordering with errors ahead of warnings so truncation can only drop the least severe; a note naming archived repos `--all` cannot cover. Removed the `checkBranchFactsParity` stub (`return nil`) and its phantom category. Package doc and help stated the locking **backwards** — Verify holds every branch read lock for its whole run and blocks all writers.

## Verification
- `go test ./...` — 44 packages green. 44 verify tests, up from 12.
- Live-home copy, `verify --all --deep`: exit 0, 12 warnings, 0 errors, 3.9s.
- `--prune-generated-refs` on a live copy: removed exactly the two `okf/*` refs and both markers; `agent/*`, `main`, and the `config` kv row untouched.
- Exit codes measured: no args → 2, unknown repo → 2, clean → 0.

A `/code-review high` pass raised 9 findings; 7 were real and are fixed in the second commit's final state (partial-prune misreporting, missing `rows.Err()` after `quick_check`, category-before-severity ordering hiding errors behind truncation, invisible marker delete, JSON `null` arrays, an ignored parameter, and a per-commit query inside the writer-blocking window). Two were accepted as documented behaviour: `--max-issues` does not truncate `--json` (truncated machine output would look complete and not be), and pruning raises the orphan-object count because knomit has no git-object GC.

## Known / not done
- **Pruning raises the orphan count** (2805 → 8378 on `knomit-kb`) since the export history becomes unreachable. Correct and expected, but a gc/repack is the natural follow-up and knomit has no such command.
- **Home-level verification of `control.db`** (registry rows with no `.db`, orphan `repo_origins`, lens rows naming dead uids, duplicate active `repo_id`) is deferred. Worth doing before `migrate-registry` is actually removed — its refusals are currently the only diagnostics for a damaged registry.

### DIFF
--- a/cmd/migrate_registry.go
+++ b/cmd/migrate_registry.go
@@ -110,6 +110,20 @@ func (o migrateOpts) out() io.Writer {
 	return o.Out
 }
 
+// deprecationNotice heads every run, including --dry-run. The command is not
+// given cobra's own Deprecated field on purpose: cobra's IsAvailableCommand
+// treats a deprecated command as unavailable and drops it from `knomit --help`,
+// and this is the command a legacy home's boot refusal tells the operator to
+// run. Hiding it would leave that refusal pointing at something they cannot
+// find.
+const deprecationNotice = `WARNING: migrate-registry is DEPRECATED and will be REMOVED in a future build.
+  It converts a home shape that predates the control.db repo registry. If this
+  home still needs converting, convert it now — once the command is gone the
+  only path off the legacy shape is re-cloning each knowledge base from its
+  remote, and a repo with no recorded remote cannot be recovered that way.
+
+`
+
 // migrateRegistryCmd builds the `knomit migrate-registry` subcommand.
 func migrateRegistryCmd() *cobra.Command {
 	var (
@@ -122,8 +136,21 @@ func migrateRegistryCmd() *cobra.Command {
 	)
 	cmd := &cobra.Command{
 		Use:   "migrate-registry",
-		Short: "Convert a pre-registry knomit home to the control.db repo registry",
-		Long: `Converts a home created before the repo registry moved into control.db.
+		Short: "DEPRECATED: convert a pre-registry knomit home to the control.db repo registry",
+		Long: `DEPRECATED — this command will be REMOVED in a future build.
+
+It converts a home shape that only exists on installations predating the
+control.db repo registry. Once your home is converted there is no reason to
+run it again, and a home created by a current build never needs it at all.
+If you still have an unconverted home, migrate it now: after this command is
+removed, the only supported path off the legacy shape will be to re-clone
+each knowledge base from its remote — and a repo whose remote was never
+recorded anywhere else cannot be recovered that way.
+
+Deliberately still listed in ` + "`knomit --help`" + ` rather than hidden: a legacy
+home makes the server refuse to boot, and the refusal names this command.
+
+Converts a home created before the repo registry moved into control.db.
 
 Run it ONCE, with the server stopped. It reads every repo database's remote
 connection config BEFORE migrating any of them (the schema migration that
@@ -148,6 +175,10 @@ creates a -shm and a zero-length -wal beside each database it opens;
 neither carries committed data and the next clean read-write close clears
 both.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			// Printed on stdout with the plan rather than only in --help,
+			// because the operator who most needs to see it is the one who
+			// pasted the command from a two-year-old runbook.
+			fmt.Fprint(cmd.OutOrStdout(), deprecationNotice)
 			if home == "" {
 				cfg, err := config.Load()
 				if err != nil {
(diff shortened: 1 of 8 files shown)


knomit:
## Second review pass — 5 findings, all fixed (`0f926e9e`)

**The first one was mine, from the previous round's fix.** Batching `commit_parents` into a single read (a performance fix from review #1) left the error path passing a nil map into the per-commit comparison. Every non-root commit then compared its real parents against nothing and was reported as drift — one transient query failure would have manufactured thousands of false errors and a permanent exit 1. That is precisely the failure this PR exists to remove, re-entering through the door the optimisation opened. The comparison is now skipped rather than run against an empty map, and a test drops the table to prove one issue comes back instead of one per commit.

The other four:

- **Flag-parse errors exited 1, not the documented 2.** Cobra returns them to `ExecuteContext`, where `main.go` prints and exits 1, so `RunE`'s exit mapping never ran — `verify --max-issues=abc` looked to CI exactly like a damaged corpus. A `FlagErrorFunc` now exits 2. The previous test asserted only the help string, so it passed while the binary contradicted it; measured before/after on the built binary.
- **`--all-branches` suppressed the generated-refs warning.** `skipped` stays nil under that flag, so `okf/*` residue moved into the checked set, produced hard parity errors, and lost the one message naming `--prune-generated-refs` — gone exactly when an operator opts in to ask why those refs have no rows.
- **Ctrl-C during `--all` printed one `context canceled` per remaining repo** as the loop opened each in turn. A cancelled context now breaks the loop.
- **`--prune-generated-refs` printed a self-contradicting report.** The pruned refs were still listed under `branches_not_indexed`, carrying the warning telling you to remove them, three lines above the output saying they had been removed. The prune now runs before the report, so the report describes the repo as it stands afterwards.

Verification: 44 packages green, 47 verify tests (up from 44). Exit codes re-measured on the binary: bad flag → 2, bad value → 2, no args → 2, clean → 0. Real-home `verify --all --deep` unchanged at 0 errors / exit 0.

PR #92, opened 2026-08-15
fix(bridge,skills): drop the misleading area TOC; teach recall the topic axis

knomit:
Two independent fixes, both for things that were quietly wrong rather than broken.

## B — the SessionStart area TOC was measuring the wrong thing

`buildAreaTOC` counted facts per area over `?sort=recent&limit=200` — **the 200 most recently committed facts, not the corpus.** The numbers read as corpus depth while actually measuring recent write activity.

The arithmetic settles it: every count in a real emitted block summed to **199** (the 200-fact window minus one excluded global principle). A quiet, well-covered area looked small; a churny one looked large; a two-month-old load-bearing invariant didn't appear at all.

It also instructed `/knomit-recall <area>` while listing ontology **topics** — a different axis (see C). Following it lands you on `applies_to: ["invariants"]`, which half-works by accident because a few facts carry `invariant` as a domain tag, and returns the wrong thing convincingly.

**Deleted, not replaced.** For a main session the CLAUDE.md knomit block already states what knomit holds and when to recall, and *"duplicated CLAUDE.md guidance"* is one of the reasons `dd5f32ba` deleted three hooks. The 200-fact fetch stays — global principles, the invariants fallback, and recent-work still use it.

The old test asserted the block **and** its per-area counts; it's replaced by one asserting their absence, so this can't return by accident.

## C — `/knomit-recall` couldn't express "the invariants for this area"

`knomit_query` has supported a `path` prefix all along; the recall skill never passed it. So the skill's own Step 1 — *"read principles first, then invariants"* — was a hope about ranking rather than something reachable.

The skill now documents both axes and when each applies:

| axis | what it matches | examples | param |
|---|---|---|---|
| subject matter | each fact's `domain:` tags | `store`, `mcp`, `repos` | `applies_to` |
| kind of knowledge | the `kb/<topic>/` folder | `invariants`, `decisions`, `gotchas` | `path` |

They combine: *"invariants about the lens write repo"* is `text: "lens write repo"` + `path: "kb/invariants/"`. Verified — `path: "kb/invariants/"` returns invariants today.

CLAUDE.md and its template carried the same `<area>` ambiguity; both now say the argument is free text and show how to name a kind.

## Scope

No Go behaviour changes beyond the deletion. Skills changed in both `.claude/skills/` and `tools/bridge/claude/templates/skills/` (verified byte-identical).

`go build ./...` clean; no failures in `./tools/...`.

## Merge note

Touches `CLAUDE-md-block.txt` on a line adjacent to one changed by #90. Both edit different lines so it should auto-merge, but worth landing one then rebasing the other.


knomit:
## KB maintenance to do after this merges

This PR rewrites `/knomit-recall`'s guidance, edits the CLAUDE.md block, and deletes `/knomit-bootstrap`. Several facts describe exactly those artifacts, so they go stale the moment this lands. Filing the list here rather than doing it on the branch, because the facts should describe `dev`, not a branch that might still change.

Already done: the deletion decision is recorded at `kb/decisions/integrations/cc-memory/delete-bootstrap-skill/1fc8ad7d.md` — options, rationale, and the two boundaries (the approve-before-writing gate was judged not-worth-a-skill rather than wrong; this doesn't generalise to the other ten skills).

### 1. Facts to update — `/knomit-update`

**`kb/decisions/integrations/cc-memory/shape-b-hybrid/703fa009.md`** — wrong in two ways, one of them pre-existing.

It enumerates the commands as *"six at the original design, now TEN: bootstrap, decided, hypothesize, principle, recall, remember, retract, review, update, why"*. That list omits `harden`, so it was already stale before this PR; and `bootstrap` is now gone. The count `TEN` survives by coincidence — the real ten after this merge are decided, harden, hypothesize, principle, recall, remember, retract, review, update, why.

Its closing sentence is now wholly wrong: *"Bootstrap is structured per-area `/knomit-bootstrap` (formerly `/knomit-kickoff-area`), then organic accumulation thereafter."* Neither command exists. Replace with organic accumulation via `/knomit-remember`, and link the deletion decision.

**`kb/conventions/integrations/cc-memory/confidence-calibration/9d5aea63.md`** — the 1.0 / 0.95 / 0.85 tiers are still correct; its anchors are not. It refs `src://knomit/tools/bridge/claude/templates/skills/knomit-bootstrap/SKILL.md`, a file this PR deletes, and lists `/knomit-kickoff-area` as an entity — a command renamed away long ago. Re-point the ref at `/knomit-remember`'s skill file (which is where the 0.85 default actually lives now) and drop the dead entity. The ref is also in the legacy `src://<name>/<path>` form with no commit or blob, so it's worth re-anchoring properly while it's open.

**`kb/decisions/integrations/cc-memory/recall-trigger-expansion/b75a1f47.md`** — lists `knomit-bootstrap` as an entity, and refs `tools/bridge/claude/templates/skills/knomit-recall/SKILL.md@b417390`. That skill has changed materially twice since: the `path`/`applies_to` axis section was added, then corrected here. The decision it records (extended triggers, note-and-proceed on empty) still holds, but "note-and-proceed on empty" is now conditional — this PR makes an unfiltered re-run a precondition of concluding "no facts". Update the body to say so, drop the entity, re-anchor the ref.

### 2. Fact to verify — `/knomit-why`

**`kb/meta/methodology/recall-discipline/7672fa0e.md`** quotes CLAUDE.md's trigger list verbatim to argue it reads as a phase rather than a habit. Two of the three bullets it quotes have been reworded (once on `dev` in #90, once here). The argument is unaffected and probably needs no edit — but it's quoting text that no longer exists, so it's worth a walk before anyone leans on it.

### 3. Facts to write — `/knomit-remember`

The three discoveries behind this PR aren't in the corpus at all. A query for the recall skill's path/domain filters returns nothing about how either one actually behaves, which is why the wrong guidance survived review in the first place.

**`path` is a raw SQL prefix, and it fails silently.** The MCP arg is passed through untouched (`parseQueryFilters`, `internal/mcp/query.go:356`) and applied as `AND f.path LIKE ?` with `q.Path+"%"` (`internal/store/search_query.go:282`). Nothing normalises it, so the ontology root must be spelled out: `path: "invariants"` matches zero rows, `path: "kb/invariants/"` matches. Measured on the live DBs — `path like 'invariants%'` returns 0 in every repo, `path like 'kb/invariants/%'` returns 146 / 13 / 82. The consequence worth recording is the failure mode, not the syntax: a malformed prefix is indistinguishable from an empty area, so it reads as "this area has no facts."

**`applies_to` is ancestor-or-equal and walks up, not down.** It maps to `DomainAncestor`, whose SQL is `domain = ? OR ? LIKE domain || '/%'` (`internal/store/search_query.go:375`). `applies_to: ["store/resolver"]` returns facts scoped to `store/resolver` *and* to `store`; `applies_to: ["store"]` misses every fact tagged `store/resolver`. Hierarchical tags are real — 17 in the largest DB — so the under-retrieval is live, not theoretical.

**The ontology topic set is per-repo, and any shipped text that enumerates it is wrong for someone.** Topics come from one of two embedded presets or a repo's own `.knomit/ontology.yaml`, and the root from `ontology_root` (default `kb`, `internal/config/config.go:202`). The two presets are nearly disjoint: `ontology_code.yaml` is invariants/architecture/conventions/decisions/gotchas/incidents/meta/principles, `ontology_default.yaml` is people/technology/science/society/culture/… — they share `meta` and nothing else. This is the root cause of the bootstrap bug: because `knomit claude init` ships skills into arbitrary repos, hardcoding the code preset's topics meant every drafted fact hit `ValidatePath`'s "unknown topic" rejection (`internal/mcp/learn.go:234` → `internal/fact/ontology.go:280`) in any default-preset repo. knomit's own repo runs the code preset, which is why it never showed here. Worth an invariant, since it's the rule that stops this recurring in the next skill.

### Not a KB item, but don't lose it

The CLAUDE.md block marker goes to `v3` in this PR. After merge, any repo carrying a `v2` block gets "its knomit block is from an older version — replace the whole block" from `claudeMdBlockNote` on the next `claude init`. That's the intended behaviour, just worth expecting rather than triaging.

PR #95, opened 2026-08-16
feat(repos,web): guided create-repo wizard, initialize mode, and the ontology invariant

knomit:
Replaces the flat three-tab create-repo form with a guided wizard, replaces the `seed` mode with `initialize`, and makes one rule true everywhere underneath: **a repository is a knomit knowledge base if and only if it has an ontology.**

## Why

The old form made you self-classify into *preset / custom / clone* before you had the information to do it. Worse, the workflow it steered you toward did not work.

`seed` required a **completely empty** remote, minted a root commit, created the consensus branch and pushed it. Every host protects the default branch on a new project, so that create failed for anyone below Maintainer:

```
seed: push main: pre-receive hook declined
GitLab: You cannot push the initial commit because the default branch is
        protected and your role does not allow it.
```

The workaround made it worse. Creating the repository with a README meant it was no longer empty, so the wizard routed to `clone` — which refused a supplied ontology and then silently substituted the default one. You picked "Code", got "General", permanently, and were never told.

## What changed

**`initialize` replaces `seed`.** It clones the chosen branch, cuts `agent/<host>` from its tip, commits the ontology there, and pushes **that branch alone**. The consensus branch is never written, so **no push access to a protected `main` is needed** — open a merge request from knomit's branch when the knowledge base should become the project's consensus. A remote with no branches is refused with the instruction that fixes it: create a `main`, one commit is enough.

`seed` was the only push to a consensus branch in the codebase and the sole reason knomit ever needed write access to one. Deleting it also removes the identity race: `initialize` takes the repo id from the remote's existing root commit, where `seed` minted a nonce, so two machines seeding one empty remote produced two identities.

**A Branch step** between Access and Ontology. Its answer — does this branch already carry `.knomit/ontology.yaml`? — decides whether an ontology step exists at all, so it cannot be folded into the origin probe, which runs before any branch is chosen. The answer is per-branch: a repo can carry the ontology on `main` and not on `develop`.

**A check that does not complete never guesses.** `initialized` has three states, and the third is not a formality: the ontology is fixed at create time, so guessing "yes" discards the ontology you chose and guessing "no" writes one over a knowledge base that already had its own. Both are unrecoverable. The step list is *derived*, so an unestablished check truncates it and there is nowhere to advance to.

**The invariant, enforced where it cannot be bypassed:**

- `loadOntology` no longer substitutes a default at any exit. A repo whose ontology cannot be established opens **read-only** — data stays reachable, and `WritableBranch` refuses every branch, so nothing is authored against a taxonomy nobody chose. `NewBindingOfLens` hardcoded `writeOK: true` and was a way in; it now derives from the same gate.
- Unknown YAML keys are warnings, not fatal, so an ontology written by a newer knomit still opens as itself. The editor stays strict — `/ontologies:validate` still reports `ok:false` for an invented key.
- Attaching an origin is gated at `ActivateSync`, which every attach path calls, as well as at the handlers that can answer with a 409 naming both ontologies.
- `CreatePreflight` establishes shape for **both** remote modes before the NDJSON stream opens, so the documented 409s are reachable rather than arriving as error lines inside a committed 200.

**One question asked in one place.** `store.BranchACreateReads` is the rule for which branch a create reads, called by `InitFromRemote` against the refs it fetched and by `ProbeInitialized` against the refs the remote advertises. They disagreed before, which made re-creating a repo this machine had already initialized a permanent dead end — the probe said "not a knowledge base", the wizard derived the only mode that answer permits, and the create refused it, forever.

**Nothing asserts what it did not establish.** Write access has three states; only a server refusing the credential is `denied`, where a timeout or a dropped connection used to render as "the host refused push access". The receive-pack dial is bounded (go-git connects eagerly with no context). A cancelled sync no longer overwrites the remote's last established status, while a caller *deadline* still does. Probe answers carry a fingerprint of the request that produced them, so a stale answer is structurally unusable instead of depending on every action to remember to clear it.

## New endpoint

`POST /repos:probe-initialized` — takes URL, credentials and **branch**, returns the three-state `initialized`. Separate from the origin probe because the answer is per-branch and the branch is not known when that runs. Documented in `openapi.yaml` along with the corrected 409s.

## Testing

Full Go suite green (`./...`, including `test/storytests`), `go vet` and `gofmt` clean. Web: 82 files / 1189 tests. `tsc --noEmit` clean. Lint unchanged at the pre-existing repo-wide count.

The shallow-clone cost of the branch check was measured against live GitHub and GitLab remotes: 215 ms for a one-commit repo, 822 ms worst case (1,409 blobs / 16.5 MB), now bounded by `Cfg.Git.MaxProbeBytes` (64 MiB default) with over-budget reported as *not established* rather than as an answer.

## Before merge

**One manual check still needs a human.** In the packaged desktop app, in the ontology editor: type `my-topic:` (must not become `My-topic:`) and `teh` (must not become `the`). CodeMirror edits in a `contenteditable` and the WKWebView shell autocapitalizes. The guard attributes are applied and unit-tested for presence, but a spike showed the automated check is *insensitive* to them — removing them entirely produced byte-identical results, because synthetic keystrokes never reach the macOS substitution service.

The `initialize` path has been exercised end to end against GitLab over HTTPS, including the empty-remote refusal and the re-create case. `clone` against GitHub over HTTPS has not.


knomit:
## Post-merge activities

Everything below was found during review and deliberately parked — nothing here blocks the merge. Ordered by what I'd do first.

*Updated after `seed` was replaced by `initialize` and the ontology invariant was enforced. Items the rewrite resolved are listed at the bottom so they are not re-opened by anyone reading the old list.*

### 1. Finish verifying the real workflow against real hosts

The `initialize` path has been exercised end to end against GitLab over HTTPS, including the empty-remote refusal and the re-create case. Still unexercised against a real host:

- **`clone` of a populated knowledge base** over HTTPS — the mode most users will hit second, when a colleague points them at an existing base.
- **A remote whose consensus branch is not `main`.** Resolution prefers `main`, else the remote's HEAD; a repo whose default is `master` or `trunk` takes a different path through `BranchACreateReads`.
- **SSH remotes.** Every automated test uses `file://`. The write probe's dial bound (go-git connects eagerly with no context) is the piece most worth watching, and a host-key change is the case that now correctly reports *nothing* rather than "push denied".

### 2. The WKWebView typing check, if it didn't happen pre-merge

In the packaged app's ontology editor: type `my-topic:` (must stay lowercase) and `teh` (must not autocorrect). CI genuinely cannot cover this — removing the guard attributes entirely produced byte-identical automated results, because synthetic keystrokes never reach the macOS substitution service.

### 3. Real defects, small

- **`internal/mcp/instructions.go:101`** — `node := ontology.Topics[name]` then `node.Description`, with no nil guard. Same bare-topic-key nil dereference fixed twice in `internal/fact`. Not reachable through the API (the serializer normalises a bare key before commit), but a hand-authored `.knomit/ontology.yaml` still parses to nil, and that file is user-editable git content. **Verified still present.**
- **`internal/web/handlers_ontologies.go:45`** maps *any* `io.ReadAll` error to 413, so a client disconnect mid-body reports "Ontology too large". Check for `*http.MaxBytesError` specifically. **Verified still present.**
- **`classifyProbeError`** doesn't treat `transport.ErrRepositoryNotFound` as auth-required. A host that 404s to hide private repos collapses the wizard to the source step, where the credential fields are unreachable. GitHub answers 401 so the common case is fine; self-hosted forges may not.
- **`ProbeInitialized` and `InitFromRemote` disagree about an empty `branch`.** The wizard always sends one, so this is reachable only through the raw API.

### 4. UI polish

- **The branch chips are not disabled while the check runs.** The stale answer can no longer route the wizard — answers carry a fingerprint of the request that produced them, and one that no longer matches is treated as unestablished — but clicking a different branch mid-check still looks like it did nothing.
- **`web/src/WizardStepRail.tsx`** — completed rail steps are `<button>`s with `border: 'none'` but no `font: inherit`, so they render in the browser's default button font while active and pending steps (`<div>`s) don't. **Verified still present**, one line.
- **`CreateRepoWizard.tsx`** — `handleProbe`/`handleCreate` set state after `await` with no unmount guard, inconsistent with `StepOntology`, which guards with a `cancelled` flag.
- **`probeError` isn't cleared on `SET_URL` or on a step change**, so an error raised on the access step follows you back to source and re-renders there until the next probe. **Verified still present.**
- **"Write your own" after an upload** re-seeds from the preset and discards the user's YAML with no confirmation. Related: cancelling the OS file dialog leaves the panel on "Checking…" indefinitely with the previous diagnostics erased, because `choose(UPLOAD)` clears `validation` and returns before touching `doc`/`source`, so the debounced effect never re-fires.
- **Duplicated style consts** (`label`, `input`, `hint`) are copy-pasted across the step files, including the new `StepBranch`. `web/src/manageStyles.ts` already exists and is already imported for `btn`.
- **Accessibility** — the ontology file input has no associated `<label htmlFor>`, and the preset cards carry no `aria-pressed` (the branch chips do).
- **Two validate round-trips per keystroke pause** — the editor's linter and the panel effect both POST the same document on the same 400 ms debounce.

### 5. Cost, now that it is measurable

A remote create makes more round trips than it needs to. `CreatePreflight` asks both probes, then `Create` re-asks and `InitFromRemote` fetches — and the wizard has already asked the same questions client-side moments earlier. The duplication is deliberate (a client-supplied answer is not a guarantee, and `Create` must stay authoritative when reached without preflight), but two things would cut it honestly:

- **`filter=blob:none` for the branch check.** `packp.FilterBlobNone()` and `UploadRequest.Filter` exist; neither is plumbed through `FetchOptions`, so it needs the lower-level session API plus a fallback for servers without the capability. The transfer is currently bounded by `Cfg.Git.MaxProbeBytes` rather than filtered — the guarantee is in place, the optimisation is not.
- **A short-lived server-side memo of the shape answer**, keyed by (url, credential, branch), used only by the fail-fast layer.

### 6. Docs and corpus

- **The design doc is now partly wrong** and should be amended rather than left as a record of what we believed: `.claude/plans/2026-08-15-create-repo-wizard-design.md` predates the branch step entirely and still models `seed`. `.claude/plans/2026-08-17-initializing-a-remote-kb.md` is the current design.
- **Run `/knomit-review`.** Several related facts landed at once, which is what the dedup/distill pipeline is for. Note that `/knomit-*` tools are unavailable when a session starts outside the repo root — that is worth knowing before wondering why recall never fired.
- **`internal/fact/validation.go:115`** references `validateKeys`, deleted earlier on this branch. **Verified still present.**
- **`internal/fact/diagnostics.go`** hardcodes the strings `"topics"` and `"children"`, duplicating the yaml tags. A tag rename would silently degrade line lookup to `Line: 0` rather than failing. The schema reflection test next door shows the pattern for catching exactly this.
- **`.superpowers/sdd/2026-08-15-create-repo-wizard-plan/`** holds the run's ledger, briefs and reports. Gitignored; delete once this lands.

### Resolved by the `initialize` rewrite — do not re-open

- ~~`clone` of an empty remote never pushes, so the reconcile loop never starts~~ — a ref-less remote is now refused for both remote modes before the stream opens.
- ~~`SET_BRANCH` is a dead action~~ — the branch step dispatches it.
- ~~`remote_sync.go`'s `Push` doc comment is inaccurate about `initSeed` pushing the consensus branch~~ — `initSeed` is gone, and the comment is accurate again: nothing pushes a consensus branch any more.
- ~~Seed probes twice~~ — superseded by the accounting in §5.
- ~~The wizard can route an in-flight check's answer onto a different branch~~ — answers carry the request they were made with.

PR #98, opened 2026-08-21
feat(bridge): add Antigravity support to the knomit bridge

knomit:
Extends the bridge beyond Claude Code so knomit can back an Antigravity project the same way it backs a Claude one.

## What's in here

- **Host-neutral `knomitapi` package** — client, facts, session, binding resolution, and server key are lifted out of the Claude-specific helpers so both hosts share one path to the knomit server.
- **Shared skill templates** — the ten `/knomit-…` skills move to `tools/bridge/skills/templates` and are served from a single embedded FS instead of being duplicated per host.
- **`antigravity` subcommand** — `antigravity init` scaffolds the plugin (`plugin.json`, `hooks.json`, MCP config, `AGENTS.md` rules); the pre-invocation hook binds the session to its project's knomit scope and injects corpus context at session start.
- **Docs** — the bridge README covers the Antigravity integration.

## Verification

`go build ./...` clean; `go test ./tools/bridge/...` passes (bridge, antigravity, claude, knomitapi, skills).


knomit:
## Post-merge follow-ups

Nothing blocking here — this is the list of things that need doing *after* the squash-merge lands, mostly because the squash and the review fixes moved code that the knomit corpus points at.

### 1. Two facts are pinned to a blob this branch rewrote

Both of these ref `tools/bridge/antigravity/hook_pre_invocation.go@78b0809c…:5bf98331…`:

- `kb/decisions/integrations/antigravity/v1-hook-scope/3bc999ab.md` — "Antigravity v1 ships ONE hook"
- `kb/invariants/integrations/hooks/guards-fail-closed/f0d9abe6.md` — "a once-per-session hook guard must fail CLOSED on a missing field"

Commit `78b0809c` was squashed away and now exists only on a local backup branch, and the blob itself changed with the review fixes (it is `23720a67…` at the branch tip). Both facts need `/knomit-update` with a ref at the merge commit — otherwise the anchors are unresolvable for anyone but me.

`guards-fail-closed` also needs a body edit, not just a fresh ref. Its rule still holds and the `InvocationNum *int` half is untouched, but its *second* worked example — "when that id was absent or failed the path-safety check, `markerPath` returned `""` and the code SKIPPED the marker check" — now describes code that no longer exists. `markerPath` hashes the id instead of rejecting it, so the only remaining unusable id is the empty one, which still fails closed. The illustration wants rewording so a future reader does not go looking for a path-safety check that isn't there.

### 2. Worth recording as a new fact

`locatePluginDir` matched on the working directory's *name* (`filepath.Base(cwd) == "knomit"`) and returned early, with no check that an `mcp_config.json` was actually there. Any workspace directory named `knomit` — this repo's own checkout is `.../knomit/knomit` — resolved to the workspace root and shadowed its own plugin dir, greeting the user on every conversation with "re-run `antigravity init`", advice that could not fix it. The generalisable shape: a directory-name convention is a hint, never evidence; corroborate it with a file that only the real directory has.

### 3. Housekeeping

- `backup/antigravity-bridge-presquash` (local only, never pushed) holds the one remaining copy of the 11 pre-squash commits. Keep it until the fact refs above are re-anchored, then delete it.
- No packaging change needed: `knomit-bridge` is already built and shipped by the Makefile's release targets, so the `antigravity` subcommand rides along with the existing binary.

### What changed since the review

`60ff33e0` closes the two findings — the `locatePluginDir` name-only match above, and `markerPath` rejecting any conversation id outside `[A-Za-z0-9_-]` (a hard skip, so a timestamped id like `conv_2026-08-18T10:30:00Z` would have disabled the greeting for every conversation of every user). Both have regression tests confirmed to fail against the pre-fix code. `go test ./tools/bridge/...` passes across all five packages; `go vet` and `gofmt` clean.

PR #99, opened 2026-08-21
feat(synthesize): cross-cluster restatement shortlist (motif phase 0)

knomit:
## The gap this closes

Prune **is** consolidation with a judge: `MergeEntry` replaces N facts with one
newly authored body under an explicit fidelity contract. But its judge only ever
sees facts that landed in the same cluster, so a restatement whose halves
cluster apart is judged by nothing and lives forever
(`gotchas/synthesize/prune-scope/c40d6748`).

This gives that judge a bounded, corpus-wide shortlist of the pairs it never
sees.

## How

A new **abstraction axis**: one title-only embedding per fact version, in a
`vec0` table keyed by `facts.id`. That key is content-addressed, so the cache is
watermark-incremental by construction — an edited fact is a new row with no
vector, and "rows lacking a vector" *is* the delta. No watermark column, nothing
to migrate.

Inside review sessions only:

1. **Backfill** tops up the axis under a named latency budget (~15s/session), so
   a large corpus reaches coverage across sessions instead of stalling the first
   one. Partial coverage is reported, never silent.
2. **A standing pair cache** is maintained SIMILAR_TO-style — top-K title
   neighbours per fact, delta-updated — so a corpus-wide question costs
   per-session work proportional to what changed.
3. **Selection** ranks by title cosine and takes what the corpus's own history
   says it may spend. Pairs at or above the model's calibrated dedup threshold
   are dropped (those are `mergeFacts`'s business), as are pairs this session's
   clusters already co-present.
4. **Emission** as ordinary `prune` work items. The prune prompt ships facts
   beside itself rather than interpolating them, so a two-fact item is
   shape-identical to a cluster item — no prompt, schema, or apply-path change.

## No number here claims a fact about a corpus

The operating point is whatever absolute cosine the last funded pair happens to
sit at *in this repo* — printed in the health output because it is a
fingerprint, not a setting. Identical code prints **0.783** on knomit-kb and
**0.884** on core.

Enforcement is entirely data-derived: enough shortlist pairs judged with none
resolved defunds a corpus; a defunded corpus still spends one probe slot every
fifth session, so recovery cannot latch shut; a resolution refunds it. A
declined pair is retired outright, so the selection window keeps advancing.

Every float literal in the phase-0 path is either a declared resource budget or
the test fails — see `TestConformance_NoCorpusPropertyConstants`.

## Measured on real corpora

`TestAcceptance_RealCorpusShortlist` (env-gated) against copies of two research
corpora, with the real embeddinggemma model:

| | knomit-kb (326 facts) | core (1 693 facts) |
|---|---|---|
| standing pairs | 2 449 (p99 0.697, p99.9 0.771) | 14 130 (p99 0.816, p99.9 0.869) |
| operating point | 0.783 | 0.884 |
| emitted | 1 | 8 (the judge-slot cap) |
| second run, unchanged corpus | 326/326 in **2 ms**, 0 embedder calls, 0 KNN lookups | — |

The research harness put knomit-kb's top pair at title-cos **0.784**; production
says **0.783**. That transfer is why the short-string template ships as measured
rather than as the doc path rendered it.

A flood corpus is **bounded, not suppressed**: core's high candidate rate costs
at most 8 judge slots per session, after which that corpus's own verdicts
decide.

## Tests worth knowing about

- `restatement_dynamics_test.go` — the class that matters. Five scenarios
  driving 4–6 sessions with partial backfill budgets, verdicts, and edits in
  between. Every defect found in review lived in cross-session behaviour, not in
  any single function; three of them were caught by this suite while it was
  being written.
- `restatement_conformance_test.go` — MN13 constant audit, review-pipeline-only
  reachability (package-scoped, fail-closed), no config surface, and a
  working-tree diff proving no `bridge*.go` change and MN5 intact. The diff check
  was verified to *fail* when a bridge file is touched.

## Scope

No motifs. No bridge-engine change. `review_effort_normal_test.go` is
byte-identical. The runtime paths (query/explain/learn) are untouched, enforced
by test rather than by intent.

**Known limitation:** the axis is epistemic-only, so restatements among
pragmatic facts stay unjudged. That is forced — prune's decision path does not
carry `Kind` through `mergedFact`, so a pragmatic fact reaching synthesis would
be silently rewritten as epistemic. Lifting it means fixing that first.

### DIFF
--- a/internal/embeddings/embedder.go
+++ b/internal/embeddings/embedder.go
@@ -160,6 +160,27 @@ func (e *Embedder) EmbedDocuments(ctx context.Context, titles, bodies []string) 
 	return embedInBatches(ctx, texts, e.embedBatch)
 }
 
+// EmbedShortStrings embeds bare short strings — fact titles, and from Phase 2
+// of the motif work, motif names and name+definition strings — through the
+// model's ShortStringTemplate. Never the query or document template: see the
+// descriptor's comment for the measurement that settled this.
+//
+// ctx is checked before each batch, exactly as in EmbedDocuments.
+func (e *Embedder) EmbedShortStrings(ctx context.Context, texts []string) ([][]float32, error) {
+	rendered := make([]string, len(texts))
+	for i, t := range texts {
+		rendered[i] = e.shortStringText(t)
+	}
+	return embedInBatches(ctx, rendered, e.embedBatch)
+}
+
+// shortStringText renders one short string through the model's descriptor. The
+// string fills the {content} slot; {title} is empty, which for the title-hack
+// template is not reached at all.
+func (e *Embedder) shortStringText(s string) string {
+	return fillTemplate(e.model.ShortStringTemplate, "", s)
+}
+
 // embedInBatches splits texts into docBatchSize chunks and feeds each to run,
 // abandoning the remainder as soon as ctx is done. Split out from
 // EmbedDocuments (rather than inlined) so the cancellation checkpoint — the
--- a/internal/embeddings/model.go
+++ b/internal/embeddings/model.go
@@ -57,6 +57,19 @@ type Model struct {
 	MaxTokens     int
 	QueryTemplate string
 	DocTemplate   string
+	// ShortStringTemplate renders a BARE SHORT STRING (a fact title, a motif
+	// name, a name+definition) for embedding. It is deliberately the title slot
+	// with a literal "none" body — the "title-hack".
+	//
+	// This looks wrong and is right: the task sweep recorded in
+	// .claude/plans/motif/2026-08-20-motif-summarizer-experiment.md measured the
+	// model card's own task prompts WORSE than this rendering for strings of a
+	// few words on embeddinggemma, and every operating point calibrated in that
+	// work assumes this exact string. Do not "fix" it to a task prompt without
+	// re-running the sweep. New embedding axes get their own calibrated
+	// operating points; never reuse fact-embedding thresholds
+	// (decisions/embeddings/model-dependent-thresholds-on-descriptor).
+	ShortStringTemplate string
 	// Thresholds are this model's cosine cutoffs for dedup/search/graph/reflect.
 	// They are calibrated per model against the real corpus (tools/calibrate),
 	// because each model has a different cosine distribution.
@@ -78,6 +91,8 @@ var registry = map[string]Model{
 		MaxTokens:     2048, // EmbeddingGemma max position embeddings.
 		QueryTemplate: "task: search result | query: {content}",
 		DocTemplate:   "title: {title} | text: {content}",
+		// The title-hack — see the descriptor comment above.
+		ShortStringTemplate: "title: {content} | text: none",
 		// Calibrated against the real knomit corpus (712 facts, tools/calibrate).
 		// EmbeddingGemma's cosine distribution runs markedly cooler than nomic's
 		// (distinct same-category pairs: mean 0.48 vs 0.75), so every cutoff is
@@ -105,6 +120,10 @@ var registry = map[string]Model{
 		MaxTokens:     2048, // nomic-embed-text-v1.5 context window.
 		QueryTemplate: "search_query: {content}",
 		DocTemplate:   "search_document: {content}",
+		// Not swept on nomic: its document rendering has no title slot, so the
+		// title-hack has no analogue here. Re-measure before trusting any
+		// short-string operating point under this model.
+		ShortStringTemplate: "search_document: {content}",
 		// nomic was the original default; these are the historical literals the
 		// thresholds were implicitly tuned against (== params.Defaults()).
 		Thresholds: params.Defaults(),
--- a/internal/embeddings/short_string_template_test.go
+++ b/internal/embeddings/short_string_template_test.go
@@ -1,0 +1,50 @@
+package embeddings
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestShortStringTemplate_IsTheTitleHack pins the measured rendering for bare
+// short strings (fact titles now; motif names and name+definition strings from
+// Phase 2 of the motif work).
+//
+// It looks wrong — a title-slot template with a literal "none" body — and it is
+// right: the task sweep recorded in
+// .claude/plans/motif/2026-08-20-motif-summarizer-experiment.md measured the
+// model card's own task prompts WORSE than this rendering for strings of a few
+// words on embeddinggemma. Every operating point calibrated in that work
+// assumes this exact string, so changing it invalidates the calibration rather
+// than merely perturbing it.
+func TestShortStringTemplate_IsTheTitleHack(t *testing.T) {
+	m, err := Lookup("embeddinggemma")
+	require.NoError(t, err)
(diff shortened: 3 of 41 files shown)


knomit:
## Post-merge: knomit facts to record

The roadmap's phase workflow defers fact-recording to after the final
implemented phase, so nothing here is written yet. This is the list to work
from when it is — either after the §8 gate, or sooner if the gate stops the
motif work after Phase 2, which would make Phase 0 effectively terminal.

Ordered by how much they would cost someone who did not know them.

### New facts

**`invariants/synthesize/shortlist-coverage-before-covered`**
A cache-state entry means "scanned against the COMPLETE corpus", never merely
"visited". Marking a fact covered while the axis is still filling freezes its
neighbours out permanently, because the cache state is exactly what says
"already covered" — and a fact scanned against a half-filled axis cannot see
partners embedded later, so the hole is silent. Refs: `restatement.go`
(`refreshRestatementShortlist`), `TestDynamics_PartialBackfillClosesOverSessions`.

**`gotchas/synthesize/knn-asymmetry-in-delta-caches`**
A cached pair may exist only because B's top-K included A, never the reverse.
Any delta maintenance that drops pairs touching a changed fact and rescans only
that fact destroys discoveries nothing will rediscover. The repair (requeue the
partners) reproduces the same bug one level out if it also deletes the
partners' existing pairs. Refs: `restatement.go`,
`TestDynamics_PairsSurviveAnUnrelatedEdit`.

**`gotchas/synthesize/throttle-without-emission-is-a-latch`**
A mechanism that defunds itself on bad outcomes cannot recover if being
defunded also stops it generating outcomes: "a resolution restores it" is
unreachable by construction. Any kill-switch keyed on its own results needs a
bounded probe path — and the probe must be consumed only when it actually
produced evidence. Generalises beyond this phase; the §3 vocabulary kill switch
has the same shape. Refs: `restatement.go` (`probeAllowed`),
`TestDynamics_DefundAndProbeRecoveryOverSessions`,
`TestDynamics_ProbeIsNotConsumedWhenNothingIsEmitted`.

**`principles/anti-patterns/corpus-property-constants`** (designer ruling, MN13)
Constants encoding a CORPUS PROPERTY (absolute cosines, expected rates) are
forbidden as fixed values — derive them from the repo's own distributions or do
not have them. Constants encoding a RESOURCE BUDGET (latency, judge slots,
structural K, SQL parameters) are legitimate but must be named and documented as
budgets where they are defined. Numbers measured on research corpora are
calibration evidence, never shipped values. Enforced by test, not by review:
`TestConformance_NoCorpusPropertyConstants` parses the phase's files and fails on
any undeclared float literal.

**`architecture/store/abstraction-axis`**
Title-only embeddings in a second code-managed `vec0` table keyed by `facts.id`.
Because that key is content-addressed, the cache is watermark-incremental by
construction — an edited fact is a new row with no vector, so "rows lacking a
vector" IS the delta, with no watermark column and nothing to migrate. Drifts
with `facts_vec` on embedding-identity change, taking the caches derived from it.
Epistemic-only, and not by preference: prune's decision path does not carry
`Kind` through `mergedFact`, so a pragmatic fact reaching synthesis would be
silently rewritten as epistemic.

**`decisions/synthesize/consolidation-scope-fix`**
Cross-cluster restatement candidates reach prune's existing judge as ordinary
`prune` work items — shape-identical to cluster items, because the prompt ships
facts beside itself rather than interpolating them, so no prompt, schema, or
apply-path change was needed. Selection is a percentile of the repo's own
pair-similarity distribution; enforcement is the corpus's own judge verdicts.
Closes the gap recorded in `kb/gotchas/synthesize/prune-scope/c40d6748.md`.

**`decisions/synthesize/verdicts-are-resolved-not-merged`**
A judge that consolidates a restatement by RETRACTING the redundant half has
done the work the mechanism exists to buy. Counting only merges defunds a corpus
that is succeeding by another route — a failure indistinguishable, from outside,
from "the shortlist finds nothing". A confidence update is not a resolution:
both facts still stand.

**`gotchas/store/sql-parameter-limit`**
SQLite caps bound parameters per statement — 32,766 on this build, 999 on older
ones. Any query building one placeholder per fact id breaks on a
whole-corpus operation, and in a review pipeline it breaks SOFT: the step
degrades to "nothing found" plus a health line. Measure the limit, do not assume
it. Refs: `abstraction.go` (`sqlIDChunk`, `chunkIDs`),
`TestAbstraction_HandlesIdListsBeyondTheSQLVariableLimit`.

**`gotchas/store/vec0-k-window-precedes-where`**
`vec0` applies its `k` window BEFORE any WHERE clause, so a KNN asking for
exactly k and then filtering (to live facts, to one branch, to one kind) comes
back short — or empty on a repo carrying many superseded revisions. Over-fetch
by a documented factor. Already implicit in `RelevantMethodologyForFact`; this
makes it findable. Refs: `abstraction.go` (`titleKNNOverfetch`).

**`meta/methodology/multi-session-dynamics-testing`**
Components that are unit-correct can fail as a system on the second session, or
the fifth, or the one after an edit — budget scaled by the wrong quantity,
caches frozen by a premature marker, a throttle that latches. Test the
SEQUENCE: several sessions with partial budgets, verdicts, and edits in between.
Every blocking defect in this phase's review lived here, and three more were
found by the suite while it was being written.

**`meta/methodology/repair-recursion`**
When a fix lands, ask what it does to its neighbours — the same class of bug
frequently reappears one level out (requeuing partners destroyed the partners'
own pairs; not-marking-uncovered still lost pairs scanned against a partial
axis; deleting a declined pair let a later rescan re-mint it). Corollary learned
the hard way: this applies to CLEANUPS as much as fixes. Two of three
self-review findings were in changes made because a reviewer called them
non-blocking tidying, which is exactly the category that gets less scrutiny.

**`conventions/store/applied-migrations-are-immutable`** (designer ruling)
An applied migration becomes immutable the moment anything outside your control
might have run it. In-place edits are legitimate only while that set is
PROVABLY empty — established by checking real databases' `schema_migrations`,
not by assuming the branch was local.

### Updates to existing facts

- **`kb/gotchas/synthesize/prune-scope/c40d6748.md`** — the gap is now
  addressed, not merely known. Add the fix ref and the residual: the shortlist
  is epistemic-only, so restatements among pragmatic facts remain unjudged by
  anything.
- **`kb/gotchas/embeddings/short-strings/1d564e96.md`** — add the title-hack as
  the shipped remedy on embeddinggemma (`title: {content} | text: none`,
  `Model.ShortStringTemplate`), with the measured consequence: the research
  operating point transferred to production within 0.001 (0.784 → 0.783)
  BECAUSE the rendering matched. It looks wrong and is right; not swept on
  nomic.
- **`kb/architecture/store/f6db3a49.md`** — the "NOT ALWAYS" clause gains a
  second worked example: a new derived table whose empty state is CORRECT ships
  without a `GraphSchemaVersion` bump, and a runtime-width `vec0` table must be
  validated on every path rather than behind the embedding-identity
  short-circuit (an upgraded DB bootstraps it at the default width, and the
  identity check then matches).

### Deliberately not recorded

The roadmap and blueprint already hold the motif design; facts restating them
would duplicate documents that are themselves canonical. Nothing about Phases
1–4 goes in until that code lands. Per repo convention the refs above use
`src://`/repo-relative forms only — no local paths.

PR #104, opened 2026-08-23
fix(synthesize): dedup merge must not re-judge refs both facts already carried

### LINKED ISSUE #103: Review session aborts entirely when a dedup cluster contains a fact citing a retracted fact (dedup gate omits prior refs)
## Symptom

On corpora where facts cite other facts that were later retracted or superseded, `knomit_review` / `StartSession` can abort **before planning anything** — every pass in the run is lost, not just the affected merge:

```
error: start session: review: dedup cluster N: dedupCluster: refs for winner
"kb/...": unresolvable fact references — nothing was written:
  ... cites kb/..., which does not exist
```

Measured on a real corpus (1,688 live facts): 21 facts (1.3%) carry refs to 25 targets that no longer resolve in the live index; a single one of them landing in a dedup cluster kills the entire session, permanently — the corpus can never run review again until the facts are hand-edited.

## Root cause

`internal/refs`' own package doc states the contract:

> A ref is checked ONCE, against the commit the write lands on. Refs a fact ALREADY carried are never re-checked ... Callers pass those as `prior`. Without this, editing a fact's title re-litigates citations written months earlier against today's corpus, and a retraction anywhere in history makes every fact that ever cited it uneditable.

The dedup merge path violates this contract. It grafts the losers' refs onto the winner and then validates with `prior = nil`, so refs the losers have carried for months are treated as newly added and re-litigated against today's live index:

- `internal/synthesize/decision.go:214` — `mergeGate.Apply(ctx, merged.Path(), mf.Refs, nil)`
- `internal/synthesize/decision.go:324` — `gate.Apply(ctx, f.Path(), df.Refs, nil)`

The reflect path at `decision.go:522` passes `priorRefsByPath` correctly — dedup is the outlier, and the codebase already contains the correct pattern.

Note the refs in question are usually **temporally valid**: the cited facts exist in the object store at the commit the referrer pinned (retracted/superseded later). Only a live-index check calls them missing. In one measured case, 454 of 479 "missing" targets resolved through `FactQuery.FactExistsAt` / the provenance walk.

## Reproduction

1. Fresh repo. Write fact B, then fact A with `refs: [<path of B>]`.
2. Retract B.
3. Write fact A2 as a near-duplicate of A (same claim reworded, cosine above the dedup threshold), so A and A2 form a dedup cluster with A2's content winning.
4. Run a review session at any effort.
5. The mechanical dedup merge grafts A's refs onto the winner, `mergeGate.Apply` receives them with `prior = nil`, the ref check fails against the live index, and the session aborts with "unresolvable fact references — nothing was written".

## Fix direction

Pass the cluster members' pre-existing refs as `prior` in both dedup call sites, mirroring `decision.go:522`. The merge then succeeds and provenance survives. Do **not** fix by dropping unresolvable refs — they are typically temporally valid, and dropping them destroys provenance to satisfy a check that should never have run on them.

## Severity

One stale citation anywhere in one cluster denies review service to the whole corpus. Mature corpora with retraction history are the most exposed.

knomit:
Fixes #103.

## The bug

`dedupCluster` grafts the loser's refs onto the winner and then calls the ref gate with `prior` set to the loser's path alone:

```go
mergedRefs := fact.UnionStrings(fullWinner.Refs, fullLoser.Refs)
mergedRefs = fact.AppendUnique(mergedRefs, loserFact.File)
gate.Apply(ctx, winnerFact.File, mergedRefs, []string{loserFact.File})
```

Every citation the two facts had carried from their own commits is therefore treated as newly added and re-litigated against today's live index — precisely what `internal/refs`' temporal contract forbids: *"a retraction anywhere in history makes every fact that ever cited it uneditable."*

`dedupCluster` returns the error rather than warning and skipping, so one stale citation anywhere in one cluster aborted the whole review session before it planned anything — and aborted again on every retry. A mature corpus could never be reviewed until the offending facts were hand-edited.

## The fix

The merge introduces no new citation — everything in the union was carried by the winner, carried by the loser, or is the loser's own path — so the whole union goes in as `prior`. The gate call stays for `Canonicalize`, and so this write path still meets the gate if a genuinely new ref is ever introduced here.

Not fixed by dropping the unresolvable refs: those targets are usually still reachable through the commit their referrer pinned, and dropping them destroys provenance to satisfy a check that should never have run.

## Scope note

The issue's fix direction names `decision.go:214` and `:324` as the dedup call sites. Those are the prune-merge and distill paths, and their `prior = nil` is correct: both write **new** facts whose refs are wholly LLM-authored. `factForLLM` carries no `Refs` field, so the model never sees the cluster members' citations and cannot copy them forward. They also warn-and-skip rather than abort, so neither produces the reported symptom. The single site that grafts pre-existing refs — and the one that aborts — is `dedup.go`, and it is the only one changed. `internal/mcp/learn.go`'s dedup-on-learn merge already passes the existing fact's refs as `prior`.

## Test

`TestDedupCluster_KeepsUnresolvableCarriedRefs` seeds two near-duplicates, each carrying a ref to a fact that does not resolve, forces them into a dedup pair, and asserts the merge succeeds with both stale refs and the subsumed fact's path intact on the winner. Against the pre-fix code it reproduces the reported error verbatim.

Full suite green.

### DIFF
--- a/internal/store/derived_from.go
+++ b/internal/store/derived_from.go
@@ -248,16 +248,21 @@ func (si *searchIndex) graphAddDerivedFromAtCommitTx(
 			return fmt.Errorf("graphAddDerivedFromAtCommitTx: resolve %s: %w", refPath, err)
 		}
 		if !ok {
-			// With the knomit_learn/knomit_update ref gate in place this is
-			// unreachable for anything written through MCP: a local ref that
-			// will not resolve is rejected before the write. Reaching here
-			// means the fact arrived another way — a direct git push to the
-			// KB branch, a remote reconcile, or a history rewrite — which is
-			// worth seeing. Skipping the edge is still correct; failing the
-			// index is not.
-			log.Warn().Str("branch", branch).Str("source", sourcePath).
+			// The ref gate stops a NEWLY ADDED local ref from being written
+			// unresolvable, but it deliberately exempts refs a fact already
+			// carried — internal/refs does not re-judge the past. So this is
+			// reachable through ordinary MCP writes too: a dedup merge grafts
+			// the loser's carried refs onto the winner, and on a corpus that
+			// predates the gate, or whose targets fell past the walk-back
+			// horizon, those refs land here on every index sync. That is
+			// expected, not a bypass — hence Debug, not Warn. A fact that
+			// arrived another way (a direct git push to the KB branch, a
+			// remote reconcile, a history rewrite) also lands here and is
+			// indistinguishable at this point. Skipping the edge is still
+			// correct; failing the index is not.
+			log.Debug().Str("branch", branch).Str("source", sourcePath).
 				Str("ref", refPath).Str("source_commit", sourceCommit).
-				Msg("derived_from: local ref did not resolve; edge skipped (bypassed the ref gate?)")
+				Msg("derived_from: local ref did not resolve; edge skipped")
 			continue
 		}
 
--- a/internal/synthesize/dedup.go
+++ b/internal/synthesize/dedup.go
@@ -233,10 +233,30 @@ func dedupCluster(
 		mergedRefs := fact.UnionStrings(fullWinner.Refs, fullLoser.Refs)
 		mergedRefs = fact.AppendUnique(mergedRefs, loserFact.File)
 
-		// Same gate as every other write path. The loser is passed as retracted
-		// because it is deleted immediately below and the winner cites it as
-		// lineage — that citation is the record of the merge, not a dead ref.
-		canonRefs, _, gerr := gate.Apply(ctx, winnerFact.File, mergedRefs, []string{loserFact.File})
+		// Same gate as every other write path — but this merge ADDS no
+		// citation, so the whole union goes in as prior.
+		//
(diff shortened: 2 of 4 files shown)


knomit:
Superseded by #109, which ports this fix onto the motif PR stack (#99→#100→#102→#105→#106→#109) so the whole stack merges bottom-up without conflict — the stack and this PR both touch `internal/synthesize/dedup.go`.

#109 carries this PR's fix and diagnosis forward with credit: the scope note here (the `decision.go` sites are correct as-is; the grafting/aborting site is `dedup.go`) was verified independently twice during #109's review and corrected the gate annex's §6.2. Two shapes evolved in the port: the carried-refs snapshot is derived from the operands in a pure helper rather than copied inline at the call site (position-independent, and now bound by a declaration-driven structural test), and per-merge failures warn-and-skip as this PR's second commit introduced. The reproduce test plants stale refs on both operands.

One operational note that travels with the fix (disclosed in #109's body): unblocking review on `core` puts the 61 facts of #107's class in the path of the still-unguarded prune-merge rewrite — hold review sessions on `core` until #107's guard question is decided.

Issue #135, opened 2026-08-25
Distill and discover can offer the IDENTICAL member set consecutively in one session — redundant work items manufacture redundant facts, and reinforcement on identical seeds fabricates corroboration


knomit:
From the live core drain (operator session s121, scope Sysdig). Filed at the operator's request. **Distinct from #127**: that issue is about redundant facts already in the corpus; this is about redundant WORK ITEMS manufacturing them.

## What happened

Item 717 (distill, grouped as a "regular cluster") and item 718 (discover, grouped as a "cross-cluster BRIDGE") carried the **same three Sysdig facts**, consecutively, in one session. Two pipelines with different stated grouping rationales converged on one trio, and each asked the judge to write a fact from it.

## The two failure modes this manufactures

1. **Duplicate facts with unmatchable bodies.** Had both items been answered, the corpus would hold two facts with the same content and THE SAME SEEDS but different origin (`distilled` vs `discovered`). Title-cosine dedup catches that pair only if the two independently-written bodies happen to be worded alike — which they are not, by construction. This is a #127-class duplicate that no similarity instrument would find.
2. **Fabricated corroboration via reinforcement.** The discover item also offered the pair as a REINFORCEMENT target. Reinforcement is defined as recording an INDEPENDENT derivation; identical seeds are the opposite of independent. Reinforcing would have incremented `sources` on the operator's own one-item-old fact using the very facts it was built from — #127's fabricated-corroboration failure reached by a different road.

The operator declined item 718 on both grounds (condition (c) — the fact had been written one item earlier — and the independence definition).

## Fix directions

- Session-scoped (or watermark-scoped) member-set dedup across item TYPES: two items whose seed sets are equal (or one a subset) should collapse to one, regardless of which pipeline generated each.
- Reinforcement eligibility should require seed-set disjointness (or at least non-identity) with the fact's existing sources — independence is checkable, not just declarable.
- Relation to #125: the lineage index proposed there gives the equality/subset check its data structure for free.

knomit:
**s126: the class is much bigger than filed — THREE pipelines collide, four of six items in one session on the same material, and the fix intersects #125.**

The original filing showed one distill/discover collision. One session later (scope CERT-UA, six items total):

```
item 727  PRUNE     e7a06b77 + 2c091c0c
item 730  DISTILL   e7a06b77 + 2c091c0c      ← identical set to 727
item 731  DISTILL   8f736433 (which CONTAINS e7a06b77's content)
[...]

Issue #111, opened 2026-08-25
Work items need a defer verb — for some item types a non-answer is a destructive write


knomit:
Surfaced by the 2026-08-24 bulk motif population on live knomit-kb.

A review-session work item can only be answered or left blocking the queue. For restatement items, an empty answer (`{"decisions": []}`) is NOT neutral: it records a DECLINED verdict against the judge-outcome throttle AND permanently retires the standing pair (`DeleteRestatementPair` runs unconditionally after the verdict; recovery only via an edit re-minting the pair). A session that wants to process only a subset of item types (our case: motif items only) therefore cannot pass through safely — blind draining of ~59 sessions would have defunded the corpus's restatement shortlist as a process artifact and retired ~59 computed candidates unjudged. We worked around it by widening the session's mandate to judge restatement items honestly.

Ordinary cluster prune items are exempt by design (`resolveShortlistPair` returns nil for non-`restate-` keys, no verdict recorded) — the gap is item types where non-answer has side effects.

Enhancement: a defer/skip verb that consumes the item for THIS session without verdicts, retirement, or any write — level-triggered return in a later session, like an unanswered item, but without blocking the queue.

knomit:
**Triage 2026-08-25 — VALID, P3, and BLOCKED-BY #120 until that ships (it now ships early in P2).**

Confirmed at `b5778535`: `DeleteRestatementPair` (`internal/synthesize/restatement.go:964`) runs **unconditionally** after the verdict, and the comment above it argues the case deliberately — a declined pair must not keep its place at the top of the ranking, or the selection window fills with pairs the judge already refused. So the retirement is not an oversight, and this issue is not a bug fix: it is a **new verb** that must short-circuit **before** `resolveShortlistPair` (`review_strategy.go:761`) records anything at all — no verdict, no retirement, no write.

The exemption for ordinary cluster prune items is real and stays: `resolveShortlistPair` returns nil for non-`restate-` keys, so no verdict is recorded. The gap is only item types where a non-answer has side effects.

## Dependency on #120

An agent cannot defer what it cannot identify. Without `ClusterKey` on the wire (#120), a defer verb ships with no way to aim it — every 2-fact prune item looks alike. #120 is now scheduled early in P2 precisely so this arrives unblocked.

## Design questions for when this reaches P3

- Does a deferred item return in a later session (level-triggered, like an unanswered item), or is it consumed for the session only? The body asks for the former; the queue semantics need stating.
[...]

PR #138, opened 2026-08-25
fix(synthesize): carry the work item's cluster key to the wire (#120)

### LINKED ISSUE #120: Work items don't carry ClusterKey on the wire — a restatement pair is indistinguishable from an ordinary 2-fact prune cluster
Measured on knomit-kb s49 (2026-08-25). The review work item exposes id/type/prompt/response_schema/facts — not ClusterKey. Consequence: an answering agent cannot follow the rule 'judge restate- items honestly, never no-op them' because it cannot tell which 2-fact prune item IS one. Concrete instance: a session whose health said 'restatement candidates emitted: 1' served two 2-fact prune items; the operator no-op'd both under the cluster rule; if either carried a restate- key, that no-op recorded Resolved:false AND permanently retired the pair (the #111 destructive-non-answer, now unavoidable rather than avoidable). Elimination from counts failed (prune/distill counts coincided but content pairing was not 1:1).

Fix: expose the cluster key (or an is_restatement flag) on the wire. Until then the only safe fleet rule is 'treat every 2-fact prune as a possible restatement pair', which wastes honest judgment on ordinary clusters.

knomit:
Closes #120. **Stacked on #137** (`fix/bridging-line-scope-local`). Review that first; this PR targets it, so its diff shows only this fix.

Also unblocks **#111** — a defer verb an agent cannot aim is not a fix.

## The bug

The work item exposed `id`/`type`/`prompt`/`response_schema`/`facts` — not `ClusterKey`. So an answering agent could not follow the rule *"judge `restate-` items honestly, never no-op them"*, because nothing on the wire told it which 2-fact prune item was one.

**That is destructive, not merely inconvenient.** A no-op on a restatement item records a DECLINED verdict against the judge-outcome throttle **and permanently retires the standing pair** — `DeleteRestatementPair` runs unconditionally after the verdict, recoverable only by an edit that re-mints it.

Measured: a session whose health read `restatement candidates emitted: 1` served two 2-fact prune items, and the operator no-op'd both under the ordinary-cluster rule. Whichever carried the `restate-` key, that no-op was #111's destructive non-answer — **unavoidable rather than avoidable**.

Until now the only safe fleet rule was "treat every 2-fact prune as a possible restatement pair", spending honest judgment on ordinary clusters every session.

## The design choice, and its cost

**The raw key, not a derived `is_restatement` flag.** The key already exists and is already computed, so it cannot drift from its source the way a second declaration would — the same reasoning as #126's schema-derived allowlist. It is also general: it identifies provenance for *every* item type, which #135's cross-pipeline collision work will want.

**The cost is real and is paid for explicitly.** Consumers now key off the `restate-` **prefix**, and a prefix nothing pins is a silent-breakage contract: change the constant and every agent's rule stops matching, with no failure anywhere. So the prefix is pinned by a test that names it as a wire contract and states what changing it costs — *"if it must change, the change is a coordinated wire change, not a rename."*

## ⚠️ Reviewer: two things to check hardest

**1. Both `PipelineItem` construction sites.** `pipeline.go` builds one in `renderWorkItem` and another in `payloadResult` (pages after the first). Populating one would leave a paged item identifiable on page 1 and anonymous afterwards — **worse than uniformly absent**, because the agent reads the later pages last, and that is when it decides how to answer. The paging test walks every page of a genuinely multi-page item and names the page that dropped the key.

**2. The engine/wire split — I got this wrong first and the sabotage caught it.**

The first test was named `...CarriesClusterKeyOnTheWire` and asserted on `PipelineItem` — the **engine** type. `ReviewItem` is what an MCP client receives, and `reviewResult` projects one onto the other field by field. **A fix that populated the engine type and forgot the projection would have satisfied that assertion completely while leaving the wire exactly as broken.**

The test now asserts **both layers, separately labelled**. Verify that the wire-layer assertion genuinely goes through `reviewResult` rather than re-reading the engine struct — that is the whole point of it.

## Sabotage evidence

Three sabotages, three distinct catches — the sites are independent and each is named by the failure.

| sabotage | caught by |
|---|---|
| `ClusterKey` dropped from **`payloadResult`** (page-2 site) | **only** `TestPageItem_CarriesClusterKeyOnEveryPage`, failing with *"page 2 dropped the cluster key"*; the `nextItem` test still passed |
| `ClusterKey` dropped from **`renderWorkItem`** (page-1 / normal site) | both tests — `nextItem` on the engine assertion, paging on *"page 1 dropped..."* |
| `ClusterKey` dropped from **`reviewResult`** (the MCP projection) | **only the WIRE-layer assertion**, by name; the engine-layer assertion in the same test passed |

That third row is the one worth reading twice: engine and wire are separate populations, and only a separately-labelled assertion distinguishes them.

## Attestations

- **MN5** — `internal/synthesize/review_effort_normal_test.go` byte-identical to `dev`.
- **MN13** — no constant added or changed. `restatementClusterKeyPrefix` is *pinned* at its existing value, not introduced.
- Full suite: 49 packages ok. `go vet` clean. `gofmt` clean.

### DIFF
--- a/internal/synthesize/cluster_key_wire_test.go
+++ b/internal/synthesize/cluster_key_wire_test.go
@@ -1,0 +1,136 @@
+package synthesize
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"knomit/internal/store"
+)
+
+// knomit#120. The work item exposes id/type/prompt/response_schema/facts — not
+// ClusterKey. So an answering agent cannot follow the rule "judge restate-
+// items honestly, never no-op them", because it cannot tell WHICH 2-fact prune
+// item is one.
+//
+// The consequence is not theoretical and it is destructive. A no-op on a
+// restatement item records a DECLINED verdict against the judge-outcome
+// throttle AND permanently retires the standing pair — DeleteRestatementPair
+// runs unconditionally after the verdict. Measured: a session whose health said
+// "restatement candidates emitted: 1" served two 2-fact prune items; the
+// operator no-op'd both under the ordinary-cluster rule. If either carried a
+// restate- key, that no-op was #111's destructive non-answer — unavoidable
+// rather than avoidable, because nothing on the wire could distinguish them.
+//
+// Until this ships, the only safe fleet rule is "treat every 2-fact prune as a
+// possible restatement pair", which spends honest judgment on ordinary
+// clusters every session.
+func TestNextItem_CarriesClusterKeyOnTheWire(t *testing.T) {
+	ctx := context.Background()
+	r, svc := newPhaseTestReviewer(t)
+	sess := manualSession(t, svc, "agent/test")
+
+	// Two 2-fact prune items, indistinguishable on every field the wire
+	// carried before this fix: same type, same shape, same fact count.
+	require.NoError(t, svc.Pipeline().InsertPipelineWorkItem(ctx, store.PipelineWorkItem{
+		SessionID:  sess.ID,
+		StepType:   "prune",
+		ClusterKey: restatementClusterKeyPrefix + "0",
+		FactsJSON:  `[{"file":"kb/technology/a.md"},{"file":"kb/technology/b.md"}]`,
+		Priority:   restatementPriority,
+	}))
+	require.NoError(t, svc.Pipeline().InsertPipelineWorkItem(ctx, store.PipelineWorkItem{
+		SessionID:  sess.ID,
+		StepType:   "prune",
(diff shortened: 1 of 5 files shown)


knomit:
The ClusterKey-on-the-wire fix from this PR (commit `e4a223f2`) **did land on `dev`** — it merged as part of PR #139, whose history contains it. This PR shows *Closed* rather than *Merged* only because a `--delete-branch` on PR #137 (the base of this one) removed this PR's base branch and GitHub closed it instead of retargeting. Tracking issue #120 is closed with the landing commit. No further action needed here.

Issue #136, opened 2026-08-25
Port the drain-earned judging discipline into the shipped prompts — independent runs currently reproduce the rails, not the judgment


knomit:
> **Revision 2** — amended after the drain operator's review (see comment below for the delta). Four rules corrected, acceptance criterion rebuilt with positive cases.

## Why

The core drain (135+ sessions, ~1,080 facts judged) produced consistently high-quality motif judging — but most of that quality is carried by the operator's session card and an authority feedback loop, not by the product. The shipped surface (item prompts, vocabulary block, MCP instructions, SHIP prose) carries the **mechanical rails**: motif format, blind definitions, df display, alias default-NO framing, discover conditions (a)/(b)/(c), schemas. It does not carry the **judging discipline**, which was earned error-by-error during the campaign and exists only in one session's card.

We have a control group: the campaign's own early sessions and incident record (#118's silently-discarded mint, #121's argument drift, the #119/#116 health-line misreadings that fooled two successive well-briefed operators) show what prompts-only behavior looks like — correct-but-shallow, making exactly the name-matching, title-pairing, and topic-collapse errors the carded rules now prevent. Independent runs (cron jobs, fresh agents, other repos) should reproduce the discipline without a hand-maintained card.

Nearly every rule below traces to a specific misjudgment that was caught, corrected, and re-tested across later sessions. The rule set is small; each entry names its target surface.

## A. Portable verbatim — backfill/judge item prompt text

1. **Read the body before the vocabulary list. No exemptions.** The residue/summary is a suggestion, not an answer. There is deliberately no carve-out for "obviously event-only" titles: deciding a title is event-only is itself a title-based judgment — the exact move this rule exists to stop. (A "plain <vendor>-flaw-gives-root" title in the record turned out, on body-read, to warrant declining the motif its near-identical sibling carries.) *Earned by:* a mechanism-bearing fact permanently lost to a residue-only decline (one-shot backfill).
2. **The fact's title, the residue, and the motif's name are all NAMES. Only the body licenses an assignment.** Pairing two names is how wrong assignments happen without a body ever being consulted.
3. **What the judge happens to know about the subject is also just a name.** If the load-bearing premise (e.g. "these weights are open") is not in the body, the assignment is not licensed. Corollary worth stating in the prompt for synthesis-vs-observation cases: a motif can be correct for a synthesis and wrong for every fact under it — the expansion clause is the synthesising author's written inference.
[...]

Issue #125, opened 2026-08-25
Discovery bridges pair syntheses with their own source facts — circular REINFORCE evidence


knomit:
Measured on core (2026-08-25): 4 of 4 discover bridge items since the corpus's backfill run resumed paired a synthesis with facts it was distilled from. Mechanism: a synthesis inherits its members' domain tags, so a domain-token bridge can always connect it to its own sources — at one session both seeds were in the same kb directory; at another, a 5-seed item held a synthesis plus THREE of its own sources. Consequence: the group makes REINFORCE look correct while being circular — it would raise `sources` on a fact (deliberately pinned at 0.65 with methodology caveats) using the evidence it was already built from, double-counting shared ancestry. The corpus itself has diagnosed the shape (a blast-radius fact carries a REDUNDANCY NOTE: re-derivations from a daily ingest are 'NOT independent confirmations').

Fix direction: bridge enumeration (or the discover prompt's gates) should exclude seed pairs connected by DERIVED_FROM lineage — the edges exist and the exclusion is computable at enumeration time. Note the REINFORCE write contract already forces sources:1 per bridge write for the shared-ancestry reason; this is the enumeration-side complement.

knomit:
Refinement from the core drain, s41–s50: the circular count is now **4 circular of 7 discover items** since the re-open, and the split is clean — **all four circular items were `kind=domain` bridges; both clean items (s48 "AG-UI", s50 "Rapid7") were `kind=entity`.**

Mechanism: a synthesis inherits its members' domain tags but not their entity lists, which is exactly why the circularity is domain-specific.

Fix-direction implication: lineage exclusion at enumeration (the fix this issue proposes) remains correct, but the operator's observation suggests a cheaper first cut — stop bridging on domain tokens entirely. On this corpus they carry almost no information ("AI safety", "agents" match hundreds of facts), and they account for 4 of 4 observed circular items. Both clean entity-kind items were still declined on the merits, so the entity lane's precision is doing its job.

knomit:
**Correction to my previous comment, from the same operator one session later (s51)** — posting before the domain-specific framing hardens.

[...]

PR #171, opened 2026-08-27
Port drain-earned judging discipline into shipped prompts (#136) — HELD for review

knomit:
Addresses #136. **HELD — do not merge.** Opened for the designer to review the full scope before any merge decision; an independent cold review is running in parallel.

Ports the earned JUDGMENT from the corpus-judging drain into six prompt surfaces — targeted additions in each prompt's own voice, not a rewrite (the prompts already carry strong rails). +57 lines across 6 files. Generic throughout (no drain-specific examples in the prompt text). Full port-list spec: `.claude/plans/motif/2026-08-26-136-port-list.md`.

## Per surface
- **prune** — complementary-by-design guard (don't merge two facts whose clauses exist to preserve a distinction); same-topic/different-event ≠ duplicate + declining-to-merge is an accountable verdict; a confidence downgrade doesn't correct a false claim (fix the body); set confidence from the evidence base; don't retract a correctly-dated snapshot.
- **distill** — distrust the synthesis you *want* true; don't launder a contradiction into corroboration; skip a cluster whose motif-carriers share no motif; + the motif block (option A below).
- **motif_alias** — a discriminator on two clauses of one fact proves the names distinct; one shared (possibly mislabelled) carrier is not evidence of sameness.
- **motif_define** — the causal-engine test (a motif names a mechanism, not a heuristic/topic/vuln-class); define what the subject does, not what the fact does.
- **reflect** — query before declaring a gap; pre-register the check; a moving rate is not a finding; widen only on evidence that moves a leg of the argument.
- **discover** — source-monoculture guard (single-vendor members ≠ a cross-community bridge).

## Decisions applied (all marked for your review)
- **Structural = option A:** the "what a motif is / when not to add one" judgment went into **distill + discover** (the surfaces that write motif-carrying facts). `knomit_learn` tool guidance left untouched — the MCP surface is unchanged.
- **Deferred:** the prune "re-judge motifs on merge" port was **left out** — it conflicts with MN11 (mechanical merge unions motifs to avoid data loss). Pending an MN11 reconciliation.
- **Bare-entity discover guard omitted** because #125 already rejects bare-entity bridges mechanically (`BridgeKind.Qualifies` = motif-only; entity is rank-tiebreaker) — verified; a prompt line would be redundant.

## Verification
`go test ./... -count=1` green (49 ok); `go vet ./...` clean; MN5 blob unchanged; MN1 conformance passes (the discover edit introduces no vocabulary token). **NOTE:** the ruling's per-surface **shakedown against a real-corpus copy** was NOT run (prompt-only edits, no corpus run) — flagged for your decision on whether it's wanted before merge.

### DIFF
--- a/internal/synthesize/discovery.go
+++ b/internal/synthesize/discovery.go
@@ -234,6 +234,13 @@ func renderDiscoverPrompt(payload DiscoverWorkPayload, ontologyRoot string) stri
 			fmt.Fprintf(&b, "      %s\n", firstLine(m.Body))
 		}
 	}
+	// Source-monoculture guard (#136, MIRROR). A bridge earns its place by
+	// spanning COMMUNITIES; if every member's material traces to one origin,
+	// the shared thread is that origin, not a mechanism that recurs across the
+	// corpus. This is a judgment the prompt must carry because the mechanical
+	// qualify gate reasons over motif recurrence and subject-disjointness, not
+	// over where a fact's content came from.
+	b.WriteString("\nA bridge must span COMMUNITIES, not one source's own material. If the members' primary sources all trace to a single vendor, project, or author, the shared thread is that origin — not a mechanism that recurs across the corpus — and it is not a cross-community bridge. Default to NO.\n")
 	// GATE rider 2 (designer, 2026-08-23). The far-lane demo established that
 	// condition (c) below is not answerable from this prompt at all: novelty
 	// rests on the agent's default-skip rather than on the 0.92 dedup gate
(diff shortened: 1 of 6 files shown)


knomit:
**Independent cold review: MERGE-READY for the held designer review — no findings.**

- Per-surface faithful to the port list; nothing added beyond it; all additions generic (no drain-specific examples in prompt text — the only drain/#136 citations are in a Go `//` comment that never reaches the model).
- Both resolutions verified: no motif-union line (MN11-safe — the distill motif block governs LLM authoring, not the mechanical union path); bare-entity correctly omitted because `BridgeKind.Qualifies` is motif-only and entity is rank-tiebreaker (verified in bridge_motif.go), while source-monoculture is a *real* gap the qualify gate doesn't cover.
- No contradiction with existing prompt instructions; none of the 'explicitly NOT ported' fixed-bug workarounds reappear.
- `go test ./... -count=1` green (49 ok); `go vet` clean; **MN5** blob unchanged; **MN1** conformance passes (discover line carries no vocabulary token); discover prompt-content tests still pass (guard sits in the shared path, outside every asserted region).

**Three LOW, non-blocking notes:** reflect's 'pre-register the check' is aspirational (outcomes are already rendered above it); discover's source-monoculture asks the agent to infer member origin from title+first-line (sound, but watch it in shakedown); the new discover line has no presence-test (consistent with the other five surfaces).

**Shakedown recommendation:** for the designer's held READ, prose+tests suffices. Before actual MERGE, the reviewer would hold the ruling's shakedown for at least **PRUNE** (highest blast radius — watch the restatement/verdict-outcome throttle) and **DISTILL** against a real-corpus copy; the thin surfaces (motif_alias, discover, motif_define, reflect) can ride on the validated protocol. Still HELD — do not merge.

PR #179, opened 2026-08-29
fix: stale live head after a long index run, and lens fact sub-resources (#178)

### LINKED ISSUE #178: Stale live head in UI after long post-merge index run (dropped SSE head broadcasts) + lens routes missing /incoming|/outgoing
## Symptom (observed 2026-08-29, repo `knomit-kb`, branch `agent/h1v302-8215ac8f`)

While browsing **live** (not in history/timeline mode), the UI showed:

```
Error: no fact at path "kb/meta/methodology/review/agreement-blind-spot/185e1fa6.md" on branch "agent/h1v302-8215ac8f" at commit "fe30076a1d3ce0addeceac83121191345a3c7ebf"
```

for a fact that exists at HEAD (reads 200 on the branch route). `fe30076a` is the `merge: main into agent/h1v302-8215ac8f (local_wins)` commit from 17:52; the fact was created at `da17d16` (18:53), after it. Additionally, no fact's in/out edges could be fetched through the lens API.

This only reproduced on the one repo that had (a) a heavy merge to index and (b) fresh local commits landing while that indexing ran. Other synced repos were unaffected.

**Ruled out during investigation** (verified against raw git objects in the store and by batch-testing the repo-scoped REST API — 30–40 facts, reads plus both edge directions, HEAD- and merge-anchored, zero failures):

- history rewrite — `mergeIntoBranchLocked` writes `ParentHashes{dstHash, srcHash}`; the agent branch stays first parent, and the raw commit objects confirm it across the GitHub PR round-trip (other agent's PR #10 → main → merged back via `local_wins`; this branch → main via PR #11)
- index corruption — `commit_parents`, `commit_log`, `branch_commits`, `branch_facts` all cross-check against the git objects
- resolver bugs — the server answers correctly for whatever anchor it is given; the wrong **anchor** was the problem

## Defect 1 — `commitObserver.Notify` silently drops head notifications (root cause)

`internal/repos/observer.go:56`:

```go
if hash != "" && o.running.CompareAndSwap(false, true) {
    defer o.running.Store(false)
    o.fn(hash)
}
```

When the 1s debounce timer fires while the previous callback is **still running**, the CAS fails and the notification is discarded — `o.pending` was already cleared before the CAS, and nothing re-arms or re-Notifies. The doc comment ("fn will not be invoked until the current run finishes") describes deferral, but the code implements a drop.

The callback (`internal/repos/builder.go:556`) is heavyweight: `Acquire → IndexManager().SyncLocked → hub.broadcastStatus(hash)`. After the `local_wins` merge, `SyncLocked` ran long (indexing + embedding the merged facts, on a machine currently capped to embed batch 1), so the SSE `status` head broadcasts for the follow-on `learn` commits were dropped. Browser tabs then held a stale `state.headCommit` = the merge commit, indefinitely. The store index itself stays correct (the inline `im.Sync` in `notifyCommit` is unaffected) — only the broadcast is lost, which is why the server looked healthy while the UI erred.

**Fix direction:** never discard — if the timer fires while `fn` is running, keep `pending` and re-arm (or re-Notify after the running callback returns) so the *last* head always broadcasts.

## Defect 2 — live-mode UI treats the cached head as authoritative and renders a raw error

In live repo context, `useFactEdges` anchors every edge fetch at the client-cached head: `edgeAnchorCommit` (`web/src/state.ts:595`) returns `state.headCommit` when live outside a lens, producing `GET /commits/{cachedHead}/facts/{path}/incoming|outgoing`. `state.headCommit` is refreshed only by the page-load status bootstrap, SSE `status` events, and the `api.status` refresh on task completion — so with Defect 1 dropping events, a live tab pins itself to an old commit with no indication.

Result: for any fact created after the stale head, the edges fetch 404s (`handlers_fact_sub.go:180` — the exact error string above) and the UI surfaces it as a raw error on a fact that loads fine at HEAD, reading as data corruption.

**Fix direction:** a 404 on a commit-anchored request while the view is LIVE should trigger a head refresh (`api.status`) and a retry, not render the error. A periodic head re-poll as a belt-and-braces fallback for lossy/dead SSE would also have masked Defect 1.

## Defect 3 — lens routes have no fact sub-resources (independent bug, same investigation)

`internal/web/router.go` registers the repo-branch route as `handleHALFact(b, p.factReader, p.factSub)` — `factSub` dispatches `/incoming`, `/outgoing`, etc. The lens route registers only `handleHALLensFact(b, p.factReader)`. A request to

```
GET /api/v1/lenses/{lens}/facts/{path}/incoming
```

matches the `facts/*` wildcard with the literal path `…/<uuid>.md/incoming` and fails with:

```
HTTP 500 — readFileWithHash: entry kb/.../<uuid>.md/incoming: object not found
```

for **every** fact, deterministically (verified live against lens `knomit-dev`). The web UI only avoids this because `useFactEdges` waits for `SET_FACT_SOURCE` and fetches edges through the source mount's repo-scoped route; any direct API consumer deriving an edge URL from a lens fact URL hits the 500, which looks like corruption but is a missing route.

**Fix direction:** register the sub-resource dispatch on the lens facts route, resolving the mount first (same resolution the lens fact read already performs), or explicitly 404 unknown sub-resources instead of misparsing them as fact paths.

## Notes

- Quick confirmation of the stale-head mechanism in an affected tab: reloading the page clears the errors (bootstrap fetches a fresh head).
- Corresponding knomit facts: `kb/incidents/ui/time-travel/sticky-pin-after-merge/0c836441.md` (corrected root cause) and `kb/gotchas/web/lens/edge-subresources/1cf58c93.md`.

knomit:
Closes #178 — three separable defects from one investigation, one commit each.

## 1 — `commitObserver` dropped head notifications (root cause)

`internal/repos/observer.go` cleared `pending` and then CAS'd on an atomic
`running` flag. A failed CAS — the previous callback still in flight — threw the
notification away: nothing re-armed, nothing re-notified. The doc comment
promised deferral; the code implemented a drop.

The callback is heavyweight (`Acquire → IndexManager().SyncLocked →
hub.broadcastStatus`), so after a `local_wins` merge it runs for many seconds
while the merged facts index and embed. Every `learn` commit landing in that
window lost its SSE `status` broadcast, and tabs held a stale `state.headCommit`
indefinitely. The store index stayed correct — the inline `im.Sync` in
`notifyCommit` is unaffected — which is why the server looked healthy while the
UI erred.

`running` now lives under the mutex with everything else, and a timer firing
mid-run leaves `pending` in place and returns; the in-flight run re-arms the
debounce timer when it finishes. The last head is always delivered. `Stop()`
keeps its semantics — flush unless `fn` is already running, so `fn` is still
never concurrent with itself.

## 2 — a LIVE view treated its cached head as authoritative

`useFactEdges` now reads a 404 in a **live** view as evidence about the anchor
rather than an answer: re-read the head, and if it actually moved, retry there
once and report the head upward so the rest of the app un-stales too. If the
head has not moved, the 404 is the real answer and is rendered — no retry loop.

Confined to live views on purpose: in history or diff mode the commit is the
user's pin, and re-anchoring it would break
`kb/invariants/ui/navigation/every-hop-is-path-plus-commit`.

`App` also re-polls the head every 30s. That is the belt-and-braces half — it
covers a stream that is merely lossy or dead, which no server fix reaches, and
it would have masked defect 1 entirely. `SET_HEAD` returns the same state for an
unchanged head, so a quiet repo re-renders nothing.

`api.explain`'s rejections now carry the HTTP status (added to a plain `Error`,
so `String(err)` is unchanged for every consumer that renders it) — without it a
404 is indistinguishable from a transport failure, and retrying the latter would
be wrong.

## 3 — lens routes had no fact sub-resources

`/lenses/{lens}/facts/{path}/incoming` matched the `facts/*` wildcard as the
literal path `…/<uuid>.md/incoming` and fell into the fact reader, which cannot
walk into a blob: HTTP 500 for every fact, deterministically. The web UI escaped
it only because `useFactEdges` fetches edges through the source mount's
repo-scoped route.

Only the lens fact handler knows how to resolve a `kb://<id12>/…` address, so
that is where the dispatch belongs. `splitFactSubResource` separates splitting
from dispatching because the lens route must strip the suffix **before**
addressing. The three sub-resource handlers now take the resolved
`*repos.RepoInstance` instead of `repos.RepoFromContext` — under `LensMiddleware`
that context value is the lens's *write* repo, the wrong mount for a qualified
path. 404s still echo the full requested path and stay byte-identical across
unknown mount / malformed `kb://` / missing fact, so the dispatch does not become
a mount-topology oracle.

## Deliberately out of scope

A fact path with some *other* trailing segment still 500s from the go-git tree
walk. That is a pre-existing wart on the repo route too, and mapping go-git's
"object not found" to a 404 would also mask genuine object-store corruption — it
needs its own decision, not a silent widening here.

## Verification

Each fix started from a test watched failing for the right reason.

- `go test ./internal/repos/` — ok (380.8s)
- `go test ./internal/web/` — ok (343.0s)
- `npx vitest run` — 1325/1326
- `go build ./...`, `go vet`, `tsc -b`, `eslint` on changed files — clean

One pre-existing failure, unrelated and unchanged by this branch:
`App.sse.test.tsx > after a resubscribe the new stream drives the app and the
old one is dead` fails identically on the unmodified tree.

https://claude.ai/code/session_01B4Fr2As3v1UZT1CiF7SE2J

### DIFF
--- a/internal/repos/observer.go
+++ b/internal/repos/observer.go
@@ -2,7 +2,6 @@ package repos
 
 import (
 	"sync"
-	"sync/atomic"
 	"time"
 )
 
@@ -14,6 +13,14 @@ import (
 // This was internal/observe, a package with exactly one consumer — this one.
 // The debounce window is a detail of how repos reacts to commits, not a
 // capability any other layer needs, so it lives here unexported.
+//
+// DEFERRAL, NEVER DISCARD. The callback is heavyweight (Acquire →
+// IndexManager().SyncLocked → hub.broadcastStatus), so after a large merge it
+// runs for many seconds. Every state transition therefore happens under mu:
+// `running` used to be an atomic the timer CAS'd, and a failed CAS threw the
+// notification away — `pending` had already been cleared and nothing re-armed,
+// so the heads of commits landing during a long run were lost and every browser
+// tab stayed pinned to a stale commit until it was reloaded (issue #178).
 type commitObserver struct {
 	mu      sync.Mutex
 	delay   time.Duration
@@ -21,7 +28,7 @@ type commitObserver struct {
 	timer   *time.Timer
 	pending string
 	stopped bool
-	running atomic.Bool // true while fn is executing; prevents re-entrant calls
+	running bool // true while fn is executing; prevents re-entrant calls
 }
 
 // newCommitObserver creates an observer that calls fn after delay has elapsed
@@ -42,39 +49,78 @@ func (o *commitObserver) Notify(hash string) {
 	}
 
 	o.pending = hash
+	o.arm()
+}
 
+// arm restarts the debounce timer. The caller holds mu.
+func (o *commitObserver) arm() {
 	if o.timer != nil {
 		o.timer.Stop()
 	}
-	o.timer = time.AfterFunc(o.delay, func() {
-		o.mu.Lock()
-		hash := o.pending
-		o.pending = ""
-		o.mu.Unlock()
+	o.timer = time.AfterFunc(o.delay, o.fire)
+}
 
-		// Skip if a previous invocation is still running.
-		if hash != "" && o.running.CompareAndSwap(false, true) {
-			defer o.running.Store(false)
-			o.fn(hash)
-		}
-	})
+// fire is the debounce timer's callback: it delivers the pending hash, unless
+// a previous run is still in flight.
+//
+// The in-flight case is the one that matters. It leaves `pending` exactly where
+// it is and returns; the run that is still going re-arms the timer when it
+// finishes, so the notification is DEFERRED, not dropped. That is what
+// guarantees the LAST head always reaches the callback.
+func (o *commitObserver) fire() {
+	o.mu.Lock()
+	if o.stopped || o.pending == "" || o.running {
+		o.mu.Unlock()
+		return
+	}
+	hash := o.pending
+	o.pending = ""
+	o.running = true
+	o.mu.Unlock()
+
+	o.fn(hash)
+
+	o.mu.Lock()
+	o.running = false
+	// Anything that arrived while fn ran is still pending — give it a fresh
+	// debounce window rather than losing it.
+	if !o.stopped && o.pending != "" {
+		o.arm()
+	}
+	o.mu.Unlock()
 }
 
 // Stop cancels any pending timer and flushes synchronously if a notification
 // was pending. Safe to call from any goroutine.
+//
+// A flush is skipped when fn is already running: that run owns the callback,
+// and flushing here would break "fn is never called concurrently with itself".
+// The in-flight run's own re-arm is disabled by `stopped`, so its pending head
+// is dropped — at shutdown there is no subscriber left to broadcast to, and the
+// index sync that matters already ran inline in notifyCommit.
 func (o *commitObserver) Stop() {
 	o.mu.Lock()
 	o.stopped = true
 	if o.timer != nil {
 		o.timer.Stop()
+		o.timer = nil
 	}
 	hash := o.pending
 	o.pending = ""
+	flush := hash != "" && !o.running
+	if flush {
+		o.running = true
+	}
 	o.mu.Unlock()
 
-	// Call fn without holding the lock.
-	if hash != "" && o.running.CompareAndSwap(false, true) {
-		defer o.running.Store(false)
-		o.fn(hash)
+	if !flush {
+		return
 	}
+
+	// Call fn without holding the lock.
+	o.fn(hash)
+
+	o.mu.Lock()
+	o.running = false
+	o.mu.Unlock()
 }
--- a/internal/repos/observer_test.go
+++ b/internal/repos/observer_test.go
@@ -55,3 +55,64 @@ func TestObserverStopFlushes(t *testing.T) {
 		t.Fatalf("expected Stop to flush pending, got %v", calls)
 	}
 }
+
+// A notification that lands while the callback is still running must still be
+// delivered. The callback is heavyweight (Acquire → SyncLocked →
+// broadcastStatus), so after a big merge it runs for many seconds; every
+// `learn` commit that lands in that window used to have its head silently
+// dropped, leaving every browser tab pinned to the merge commit until it was
+// reloaded (issue #178).
+func TestObserverDeliversNotificationArrivingDuringCallback(t *testing.T) {
+	var mu sync.Mutex
+	var calls []string
+	blockFirst := true
+
+	started := make(chan struct{})
+	release := make(chan struct{})
+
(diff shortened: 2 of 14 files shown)


knomit:
Second round: a high-effort review of this branch turned up three findings, all reproduced before fixing, now addressed in `1a8831f6`.

**HIGH — the new lens sub-resource dispatch was a mount-topology oracle.** The dispatch runs after mount resolution, and from there the handlers report failure through `writeStoreError`, never `lensFactNotFound` — and `handleFactCommits` doesn't fail at all for a path with no commits, it answers an empty **200** whose `_links.self` names the mount's repo *and* branch:

```
MOUNTED id, /commits   -> 200 {"count":0,"_links":{"self":{"href":
                               "/api/v1/repos/beta/branches/machine:test/facts/kb/y/2.md/commits"}}}
UNMOUNTED id, /commits -> 404 {"detail":"no fact at path \"…/commits\""}
```

So a caller could iterate candidate `id12`s and read "mounted" off 200-versus-404. That contradicted `lensFactNotFound`'s own doc comment and `openapi.yaml:1728`, both of which promise byte-identical 404s. Sub-resources are now gated on the same read the fact endpoint performs.

**MEDIUM — the head poll could move the head backwards**, re-staling the tab it exists to un-stale: nothing ordered a slow poll response against the SSE stream. It now drops a response if the head moved while it was in flight, and dispatches `SET_HEAD` rather than `SET_STATUS` (the latter rebuilds state unconditionally, so the previous version re-rendered every open tab every 30s — exactly what the comment claimed it didn't do).

**LOW — `useFactEdges`' retry result is always discarded in App's wiring**: reporting the head re-anchors state, re-running the effect, whose cleanup staleness-drops the retry's own fetches. The retry stays (it's what makes the hook correct without a wired `onHead`), but the test no longer asserts a path the product doesn't take — it now also pins the App-shaped caller.

**Correction carried into the knowledge base.** The rationale I recorded for rejecting the "drop the live commit anchor" option was wrong: `graphStore.ExplainFact` doesn't serve "the current graph" — it resolves the path's HEAD-active commit and delegates to `IncomingAtCommit`/`OutgoingAtCommit`, the same version-aware primitives, so that swap would not have broken `target_commit` pinning. The choice stands on other reasons; the decision fact and the imprecise `api.ts` comment it was inherited from are both corrected.

Cleared in review and unchanged: the observer's lock discipline (all interleavings walked), the `Stop()`-while-running drop, repo-route behaviour across the sub-resource refactor, and the live-only gate on the head retry.

Verification: `go test ./internal/web` ok (341.8s), `go test ./internal/repos` ok (380.8s), `vitest` 1327/1328, `go build`/`go vet`/`tsc -b`/`eslint` clean. The one failure — `App.sse.test.tsx > after a resubscribe…` — is pre-existing on `dev`, confirmed by stashing and re-running on a clean tree.

PR #180, opened 2026-08-29
fix(web): motif UI follow-ups — motifs everywhere, a path-scoped vocabulary, the pivot as a mode, and a borderless edges row

knomit:
Follow-ups to #177, in two rounds of review against the landed motif UI. Three
commits, all client-side; no Go changes.

## Round 1 — naming, a way back, and a path scope

**Call them motifs.** `ShapesBlock.tsx` → `MotifsBlock.tsx`, every `shapes-*`
test id → `motifs-*`, and the copy with them: the pivot header, the edges-row
cell, the panel heading, the empty and loading states.

**`+N more` opens a browser, not an in-place expansion.** It borrows
`FacetBrowser`'s grammar to the letter, which is the point — the same words a
few pixels apart must not mean two different things. `← Back` replaces the
glyph, search takes focus, Escape closes, the list scrolls inside a fixed 232px
so Highlights is not shoved down the panel, back drops the query, a pick closes.

**The ontology path scopes the motif vocabulary.** New
`ClustersUnder(ctx, branch, pathPrefix)` and `VocabularyHealthUnder(...)`; the
unscoped calls delegate with `""`. A cluster no fact under the prefix carries is
absent, `DF` counts carriers in the subtree, and a new `DFTotal` keeps the
branch-wide figure — both from one pass so they cannot disagree. Members and the
elected canonical stay branch-wide, so a scoped view never renames a cluster the
pivot heading names differently. `?path=` is scope, so it narrows the health
block too; `q` is a way of reading one page and does not.

The first fix landed a second: at a folder holding one carrier the block drew
`Motifs 1` over empty space. The singleton fold was written for a whole-repo
vocabulary, where `df == 1` means "minted once and never reused"; under a path
scope it means "one fact HERE carries it". The fold now keys on the branch-wide
count, and folds only while the overflow row can reopen it — **nothing the block
holds may be unreachable, and it never shows a count with nothing under it.**

### The widen rungs are gone

`this motif 14 | stem — | token-2 —` was dead in every corpus and misleading
where it was not. Measured before removing, across all four live corpora:

| corpus | clusters | `stem` live | `token-2` live |
|---|---|---|---|
| agentic-engineering | 50 | 0 | 3 |
| knomit-kb | 50 | 0 | 3 |
| core | 50 | 0 | 9 |
| knomit-io-kb | 28 | 0 | 2 |
| **total** | **178** | **0 (0%)** | **17 (10%)** |

`stem` can only differ from exact on a judge-merged cluster, and there are zero
judge merges anywhere. The 10% where `token-2` fires is worse than the 90% where
it does not: on `control-coarser-than-failure` it admits three
`failure-faster-than-success` facts because both names contain `than` and
`failure` — two motifs about unrelated things, joined by a stopword, presented
as a fact about the corpus.

Removed with the control: `state.motifMatch`, `SET_MOTIF_MATCH`, its `NavEntry`
field and both `NAV_BACK` restores, the count-only `token-2` probe fired once per
pivot, the `admittedBy`/`near` row marker, and FilterBar's dashed widened chip.
The server's `motif_match` and the MCP tool's are untouched; the client now sends
no tier at all, so the server's default is the single definition of exact.

### The pivot is a MODE, and now says so

A motif pivot is a peer of ontology / chrono / search — not a place, not a
moment, not a ranking. It gets a `≈` segment that appears only while a pivot is
on, exactly as the relevance radar appears only while searching.

Both are **derived**: nothing in the strip can enter either, so both segments are
exits. Clicking the lit `≈` drops the motif chip and returns you to the mode you
came from, via a new `EXIT_MOTIF` twinned with `EXIT_SEARCH`.

That required `PIVOT_MOTIF` to stop writing `librarySort`. The pivot already
arrives in Recent by derivation — a motif chip is a content filter and the tree
cannot honour one — so writing it stored the derived value over the remembered
one and erased the only record of where the reader came from. An ontology
browser who pivoted could only ever be dropped back into a flat feed. It is the
same failure `EXIT_SEARCH`'s own comment describes for relevance.

### A count beside a list must be OF that list

Reported as *"adding a domain or entity chip after a pivot has no effect."* The
filtering was never broken. Two things were going on:

1. The picker lists `agentic engineering` first, and all 302 facts in that repo
   carry it — so that particular chip genuinely changes nothing.
2. **The real bug:** the pivot heading printed the cluster's `carrier_count`,
   which ignores every other chip. Adding `domain:reliability` took the rows
   14 → 8 while the number beside them went on saying 14. The one reading a
   reader checks to see whether their filter did anything was the one reading
   that could not move.

`MotifPivot.carrierCount` is deleted; the header renders the same `count` every
other mode renders. One number, one source.

### No border on the edges row

Nothing else in the fact's chrome wears one, and a box around three of the
header's cells read as a widget bolted onto the fact rather than part of it.
`controlStrip` loses its border, fill, radius and hairline dividers.

The four states the border was credited with are the cells' own and are
unchanged — verified in the browser by computed style: inert `#333`, live
accent, failed red, open `#151515` plus an inset underline in `currentColor`.
With the dividers gone the padding *is* the spacing, so `CELL_PAD_X` is shared
by every cell and the strip cancels exactly that much on its left edge, putting
the row's first glyph in the same column as the path and meta lines above it.

Also: `≈ same motif as` is gone from the pivot header — the chip and the mode
segment each carry a `≈` already, and the name does the work. The slot still
renders a non-breaking space so the header keeps its height and the list does
not jump on the first pivot.

## Verification

Driven in a browser against the live corpus, not only in tests:

- pivot → 14 rows / header 14; add `domain:reliability` → **8 rows / header 8**
  (was 8 rows / header 14).
- The `≈` segment lit, `aria-label="Leave this motif"`, path and recent disabled.
- Round trip: Ontology (7 topics) → pivot → narrow → leave → domain chip
  survives, motif chip gone → remove it → **back in the Ontology tree, 7
  topics.** Impossible before, because `PIVOT_MOTIF` overwrote `librarySort`.
- Edges row: `border: 0px none`, `background: rgba(0,0,0,0)`, `gap: 2px`,
  `margin-left: -7px`, every cell `padding: * 7px`.

The count regression test was checked by sabotage — pinning the header to a
constant fails it — because the old fixture returned the same number for the
cluster and for the list and so could not tell them apart.

**Tests:** 95 files / 1328 passing. `state.motif.test.ts` was rewritten from a
tier test into a round-trip test; `LibraryHeader.test.tsx` gains six for the new
mode.

## One thing to know before merging

**`npx tsc --noEmit -p tsconfig.json` checks nothing.** That file is
`{"files": [], "references": [...]}` — a solution file. With `-p` and no
`--build` it type-checks an empty file list and exits 0 on anything, which is
why round 1 reported "tsc clean" while errors were sitting in the tree. Use
`-p tsconfig.app.json` or `tsc -b`. Worth fixing in CI separately.


knomit:
## Post-merge: knomit corpus updates

Not to be run before this merges — the facts below should describe `dev`, not a
branch. Four existing facts have drifted and two new ones are owed.

### Update — `kb/decisions/ui/motif/pivot-and-widen/0a23fe61`

**Point 2 (the widen control) is RETRACTED. Points 1 and 3 stand unchanged.**

The fact records the rungs as a deliberate design with a measured-delta rule and
an inert-when-empty rule. Both were honoured; the control was still removed,
because the measurement it was built to display had no signal in it. Numbers
taken from all four live corpora before deciding:

| corpus | clusters | `stem` live | `token-2` live |
|---|---|---|---|
| agentic-engineering | 50 | 0 | 3 |
| knomit-kb | 50 | 0 | 3 |
| core | 50 | 0 | 9 |
| knomit-io-kb | 28 | 0 | 2 |
| **total** | **178** | **0 (0%)** | **17 (10%)** |

`stem` was dead everywhere, for the reason the fact itself gives — it can only
differ from exact on a judge-merged cluster, and there are zero judge merges in
any corpus. The 10% where `token-2` fires is worse than the 90% where it does
not: on `control-coarser-than-failure` it admits three
`failure-faster-than-success` facts because the two names share `than` and
`failure`. Two motifs about unrelated things, joined by a stopword, presented to
the reader as a fact about the corpus.

The update should keep the fact's own reasoning (the inert-rung rule was right,
and `stemCanAdd` deriving the answer for free was right) and record that the
feature died of its own measurement, not of neglect. Worth carrying as the
lesson: **a control whose value is a corpus property has to be measured against
the corpus before it ships, not only after.** The
`anti-patterns/corpus-property-constants` principle says do not hard-code a
corpus property; this is its neighbour — do not ship a control for one without
first checking the distribution is non-degenerate.

### Update — `kb/decisions/ui/motif/edges-row-header/28608045`

Describes the row as `↙ cited by N | ↗ cites N | ≈ motif-name count … [+N]` in
**one bordered strip**. The border, the fill and the hairline dividers are gone;
nothing else in the fact's chrome wears one, and a box around three of the
header's cells read as a widget bolted onto the fact.

Everything else in the fact survives and should be stated more plainly now that
the border is not carrying it:

- The three-row header, and the reopening rationale (name-controls, not
  count-readouts) — unchanged.
- Version, date and retract still sit apart from the panel-openers. What
  separates them is now the left/right split in `fact-band-edges`, not a border.
- The three-state count discipline (value / `···` / `!`, never a false 0) —
  unchanged, and now the *only* thing distinguishing the cells.

### Update — `kb/decisions/ui/connections/header-cells/7a757d22`

Same edit, from the other side. Its "SUPERSEDED IN PART (2026-08-29, PR #177)"
paragraph names "labelled cells in a bordered strip"; drop the border from that
sentence.

**Its load-bearing content is untouched and should be reaffirmed, not softened:**
ZERO IS NOT A BUTTON and A FAILED FETCH IS NOT A ZERO both still hold, and with
the border gone they are doing more work than before — colour and cursor are now
the whole of what separates inert from live. Verified in the browser after the
restyle: inert `#333`, live accent, failed red, open = `#151515` fill plus an
inset underline in `currentColor`.

The "WHAT THIS DOES NOT MEAN" list needs one more entry: *not* that the cells
are drawn inside a bordered container that owns their radius and dividers — they
are borderless cells on the band's own ground, spaced by a shared `CELL_PAD_X`.

### Update — `kb/decisions/ui/motif/vocabulary-fourth-block/b5534397`

Round-1 debt, still outstanding. Calls the block **'Shapes'** and names
`ShapesBlock.tsx`; both were renamed to Motifs. The same update should carry the
path-scope decision: `?path=` narrows the vocabulary AND its health block,
members and the elected canonical stay branch-wide, and the singleton fold keys
on the branch-wide count (`df_total ?? df`) so a scoped view never shows a count
with nothing under it.

### New decision — the motif pivot is the fourth MODE

Peer of ontology / chrono / search, not a sort axis. Three claims worth pinning:

1. **It is DERIVED, like relevance.** The motif chip is what makes the list a
   pivot; there is no stored mode. Two of the four modes therefore cannot be
   entered from the mode strip, which is why each of their segments is an EXIT —
   clicking the lit one leaves.
2. **`PIVOT_MOTIF` must not write `librarySort`.** The pivot already arrives in
   Recent by derivation (a motif chip is a content filter; the tree cannot
   honour one). Writing it stored the derived value over the remembered one and
   erased the only record of where the reader came from — the exact failure
   `EXIT_SEARCH`'s comment already describes for relevance. An ontology browser
   who pivoted could only ever be returned to a flat feed. Same bug, second
   mode, and the first one's write-up did not stop it: **a rule stated about one
   instance of a pattern does not generalise itself.**
3. **A derived mode disables the rest of the strip.** Not because it is a mode,
   but because it forces the ordering: in a pivot, Recent's only remaining
   effect is to write the axis the list already has and clear the open fact.

### New gotcha — a count beside a list must be OF that list

The pivot heading printed the cluster's `carrier_count`. That is the same number
as the list total for as long as nothing else narrows the list — which is why it
passed review, and why the mock in `Library.pivot.test.tsx` returned 26 for both
and could not tell them apart.

Add a domain chip and the rows went 14 → 8 while the number beside them went on
saying 14. Reported as *"adding a chip has no effect"*: the filtering was never
broken, but the one reading a reader checks to see whether their filter did
anything was the one reading that could not move.

The general shape, which is the part worth keeping: **two quantities that agree
in the common case are not thereby the same quantity, and a test whose fixture
makes them equal cannot tell you which one you are reading.** The regression test
now returns a narrowed total that is deliberately *not* the carrier count, and
was checked by sabotage — pinning the header to a constant fails it.

### Suggested order

`/knomit-update` the four drifted facts first (they are corrections to the
record), then `/knomit-decided` the two new ones, then `/knomit-review` — this
is a burst of fact-writing in one area, which is exactly the prune/distill case.

PR #182, opened 2026-09-01
A lens has one motif vocabulary

knomit:
Motif filtering for lenses, working the way it does for repos.

## What was missing

The list endpoints already parsed `motifs=` / `motif_match=`. What did not exist was the **vocabulary** behind them: there was no cross-mount cluster identity, so the vocabulary block, the `motif:` filter category and the pivot's cluster resolution were all dark whenever a lens was open.

## What this adds

- `GET /lenses/{lens}/motifs` and `/lenses/{lens}/motifs/{key}` — every mount's motif clusters merged into one vocabulary, in the per-repo endpoint's own envelope.
- The union itself lives in `internal/federate`, because **two** surfaces answer motif queries against a lens: the REST lens reads and the MCP `knomit_query` fan-out.
- The web client picks its surface through one `MotifEndpoint` value derived from app state, so `MotifsBlock`, `useMotifClusters`, the filter picker and the pivot heading each gained a lens without gaining a lens code path.

## How clusters merge

A `cluster_key` is a mechanical function of a spelling (canonicalize → tokenize → sort → join), so the same key names the same shape in every repo. That is not sufficient: a judge merge is **per-branch**, so one mount can file `quiet-degradation` under `fallback-silent` while another still keys it `degradation-quiet`. Merging on the key alone puts one spelling in two rows that answer different queries.

So clusters union transitively on **shared key OR shared member spelling**. The merged key is the smallest constituent one — stable under df change, which is the property definitions and URLs need.

## The semantic change to know about

**A judge merge on any mount changes what a motif term returns for the whole lens.** Read through a lens the mounts share one vocabulary, exactly as one repo's folders share one. It is **read-only**: no mount's alias table, canonical election or verdict history is touched, and the same repo read directly still answers with its own vocabulary alone. A lens over a single mount is left alone entirely, so it answers exactly as that repo does.

Two tests asserted the opposite rule and stayed green only because their fixture had no motif index, so the widening never ran. They now assert the new rule, with a companion pinning the read-only half.

## Review rounds

Three rounds, each finding something real:

1. **The pivot silently dropped mounts its own count included.** The store's exact tier is per-branch canonical equality, so a row reading `df 7` listed 5. Fixed at the endpoint (`federate.ExpandMotifTerms`) rather than in the callers, so a chip minted anywhere is correct because the endpoint resolves it.
2. **MCP was never widened**, leaving the same defect live on the most-used motif path. That is what moved the union into `federate` — one definition both fan-outs reach.
3. Smaller: the `N of M` denominator was a different population from its rows; alias attribution read the union's key set where it needed the per-mount one; the widening's error dropped the branch name from `no branch named "<branch>"`.

Every regression test here was sabotage-verified — the mutant that should kill each one does.

## Accepted trade-offs, stated

- `df` / `df_total` and the pooled health counts are per-mount sums, carrying the same fork over-count the lens `/stats` histograms accept. Documented in the OpenAPI spec.
- Because a term's meaning is lens-wide, a mount excluded by `repo=` is still read to resolve it — so that mount failing fails the request. Deliberate under RFC §9.1, and pinned by a test so it is not later tidied away as a bug.
- Over two or more mounts a term can reach spellings a single unrebuilt repo's exact tier would not. That is the shared vocabulary, not a leak in the tier.

## Verification

`go test ./...`, `tsc --noEmit`, `vitest` (1362 tests), `eslint` at one warning below baseline.

### DIFF
--- a/internal/federate/motifs.go
+++ b/internal/federate/motifs.go
@@ -1,0 +1,354 @@
+package federate
+
+import (
+	"context"
+	"sort"
+
+	"knomit/internal/repos"
+	"knomit/internal/store"
+)
+
+// The lens-wide motif vocabulary: every mount's clusters merged into one.
+//
+// A motif cluster_key is a MECHANICAL function of a spelling (store's
+// groupingKey: canonicalize → tokenize → sort → join), so the same key names
+// the same shape in every repo and merging by key is sound. It is not
+// SUFFICIENT, though: a judge merge is per-branch, so one mount can have pulled
+// `quiet-degradation` under the key `fallback-silent` while another still keys
+// it `degradation-quiet`. Merging by key alone would then emit that one
+// spelling in two rows of one vocabulary, and the two rows would answer
+// different queries from names the reader cannot tell apart.
+//
+// So per-mount clusters are unioned TRANSITIVELY: two clusters are the same
+// cluster when they share a cluster_key OR share any member spelling (within a
+// mount a spelling belongs to exactly one cluster, so a shared spelling is
+// evidence the two are one shape).
+//
+// THIS LIVES IN federate, not in a consumer, because two surfaces answer motif
+// queries against a lens — the REST /lenses/{lens}/{facts,search,motifs} and
+// the MCP knomit_query fan-out — and a second implementation kept in step is
+// exactly the failure the union exists to prevent. The web surface shipped the
+// widening first and MCP did not, which left the most-used motif path (recall)
+// still dropping mounts; one definition here is what makes that impossible to
+// repeat.
+
+// MotifGroup is one merged cluster: what every mount contributed about one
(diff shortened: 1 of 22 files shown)


knomit:
Knowledge captured from this PR into the knomit KB (knomit-kb, branch `agent/mindev.local-8ef0cd32`):

- `kb/decisions/lens/motif/cross-mount-cluster-identity/1d35c7b0.md` — A judge merge on ANY mount changes what a motif term returns for the WHOLE lens (reads only); the union merges clusters by shared key OR shared spelling
- `kb/decisions/web/ui/motif/endpoint-not-context-branch/bbdb89e7.md` — The motif UI takes a MotifEndpoint, not an isLens flag: one derivation, and no component below it knows a lens exists

PR #184, opened 2026-09-02
Remember what a word's singular is

knomit:
> Supersedes #183, which was auto-closed when its stacked base branch was deleted on the merge of #182. Content unchanged.

Stacked on #182 — base is `feat/lens-motif-filtering`, and it retargets to `dev` when that merges. One commit, 3 files, all in `internal/textnorm`.

## What was actually slow

The brief was "the lens motif feature is slow, and the suspect is the per-request union rebuild fanned out over every mount". Measured, that is not where the time went. The fan-out is honest: a lens costs **exactly the sum of its mounts**, with no superlinearity and no server-side N+1. The per-mount unit cost was the whole problem — a **single repo's** `/motifs` took ~40ms.

A stack profile under load, then an instrumented build counting calls, found this:

```
knomit_motif_key (SQL callback)
  → store.groupingKey
    → textnorm.Tokens
      → textnorm.Stem
        → go-pluralize Singular    ← backtracking regexes, ~25µs per token
```

`ClustersUnder` computes the cluster key **in SQL**, for every row whose motif has no stored alias. Nothing cached the pluralizer's answer, so that ~25µs was paid per token, per motif, **per row**, on every request. One `GET /lenses/all/motifs` over five mounts:

| mount | ClustersUnder | groupingKey calls | groupingKey ms | share |
|---|---|---|---|---|
| knomit-kb | 25.2 ms | 576 | 24.1 | 96% |
| core | 30.0 ms | 732 | 28.6 | 95% |
| knomit-io-kb | 2.4 ms | 68 | 2.2 | 90% |
| agentic-engineering | 26.4 ms | 640 | 25.1 | 95% |
| test | 0.1 ms | 0 | 0.0 | — |

**It only bites where the alias table does not cover the branch.** The `COALESCE` in `motifClusterKeyExpr` short-circuits on a stored `cluster_key`, so a fully rebuilt branch never calls `knomit_motif_key` at all. A branch with zero coverage — a freshly created agent branch, which a repo takeover makes — pays for every row; a partially covered one pays for the rest. Neither is exotic: `ClustersUnder`'s own comment says any corpus with facts written since the last rebuild is in this state.

| `knomit-kb` `/motifs` | zero alias coverage | partial coverage |
|---|---|---|
| before | 39.3 ms | 9.5 ms |
| after | 2.9 ms | 2.4 ms |

## The fix

Memoize `textnorm.Stem`.

`Stem` is a **pure function of its argument**, so remembering its answer needs no invalidation at all — it is the same answer, not derived corpus state that can go stale. That is why the memo lives on the primitive rather than as a per-branch cache in a consumer: one seam, and `groupingKey`, domain matching, the motif-term query tiers and the motif subject-word strip all get it. Free-text search is *not* among them — `opts.Text` goes to the embedder and never through `textnorm`.

Unbounded, matching `store.branchCache`'s posture. What bounds it is worth stating exactly, because it is **not only the corpus**: the keys are corpus vocabulary *plus whatever domain and motif filter tokens callers have asked about* — `?domain=` and `?motifs=` reach `Stem` through `canonicalizeDomain` and the motif tiers, so a request can mint an entry. Any size to evict at would still be a guess about how big a corpus is.

Because request strings do get in, an entry must not retain more than itself. `Tokens` hands `Stem` a `strings.FieldsFunc` substring sharing the canonicalized input's backing array, so **the key is cloned before it is stored, and the clone is what gets singularized** — `Singular` returns its argument unchanged for an already-singular word, which would otherwise pin the caller's buffer through the value too. One token off a long `?domain=` would hold that whole value forever. The clone costs one allocation on a miss; the hit path stays at 0 allocations.

## Before / after

Lens `all`, 5 real mounts, `curl` n=15. Both binaries against the same home, one at a time, port ownership checked on every run.

| endpoint | before | after | |
|---|---|---|---|
| `GET /lenses/all/motifs` | 126.8 ms | 8.6 ms | 14.7× |
| `GET /lenses/all/motifs/{key}` | 105.5 ms | 8.3 ms | 12.7× |
| `GET /lenses/all/facts?motifs=` (widening) | 106.2 ms | 8.3 ms | 12.8× |
| `GET /lenses/all/facts` (no motif filter) | 3.8 ms | 3.8 ms | — |
| `GET /lenses/all/stats` | 85.9 ms | 85.2 ms | — |

Committed benchmark, `go test ./internal/textnorm/ -run XXX -bench 'Stem|Tokens'`. **Recorded as orders, not digits** — the warm figure moves run to run, and an earlier draft of this comment recorded that noise as a result:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| `BenchmarkStemCold` | ~25 µs | ~480 | 13 |
| `BenchmarkStem` | tens of ns | 0 | 0 |
| `BenchmarkTokensCold` | ~45 µs | ~1450 | 35 |
| `BenchmarkTokens` | hundreds of ns | 128 | 2 |

The allocation counts are the part worth pinning exactly; the hit path's zero is the one a future change must not move.

**Cold path.** The memo is per process, so a restart pays it once — but boot-time index work already warms most of it. First three requests on a freshly booted server: **22.5 / 13.2 / 12.3 ms**, settling at 8.6 ms. Even the coldest request is 5.6× faster than the pre-fix steady state.

## Semantics: unchanged, and checked two ways

**Test.** `TestStemMemoIsSemanticallyTransparent` runs 45 real motif spellings through `Tokens(Canonicalize(·))` against a written-out unmemoized reference, on the cold call and the warm one. `Tokens` feeds `groupingKey`, so a memo that answered differently for even one token would silently re-key clusters — a spelling would join, or stop joining, a different cluster.

**End to end.** 17 responses captured from both binaries against the same home — the lens motif collection (both pages), a cluster detail, a motif-filtered facts read, an unfiltered facts read, a lens search, lens stats, and each of the five mounts' `/motifs` on two branches. **16 of 17 byte-for-byte identical.** The 17th, `/lenses/all/stats`, differs in one field — `changes_7d` 438 vs 439, a time-window counter, and the captures were minutes apart. No motif surface differs at all.

### The pin test needs two probes, and that is the point

Retention is asserted on `unsafe.StringData`, not on equality: the strings are equal whether or not the entry aliases its caller's buffer, so an equality assertion passes on the leaking version.

It takes **two different probe tokens**, one per half:

- **key** — `"vulnerabilities"`, a word `Singular` rewrites.
- **value** — `"padding"`, a word `Singular` leaves alone.

A rewritten word can only ever prove the key, because the value is a fresh allocation whatever the code does. Checking the value with a rewritten word *looks* like coverage and is not.

## Review provenance

Reviewed adversarially, and the review found real gaps rather than nits:

- Three prose findings, all verified against the code before being accepted: the design-rationale sentence claimed the keys were corpus vocabulary only, when `?domain=` and `?motifs=` demonstrably reach `Stem`; "search tokenization" overclaimed, since `q.Text` never touches `textnorm`; and "a few dozen bytes per entry" understated retention, which produced the `strings.Clone`.
- One coverage finding: the first pin test asserted on the key only, and its probe token made the value side structurally unreachable. The reviewer proved it with a mutant — clone the key but singularize the *original*, restoring half the leak — that left the suite green. That mutant now fails.

Every guard here is sabotage-verified. Five mutants, each killed by a named test: a memo hit returning the raw token, a memo keyed on the output, a dropped `-ics` guard, a dropped `Clone`, and singularizing the original instead of the clone.

## Verification

`go test ./...` exit 0 · `go vet` clean · `go test ./internal/textnorm/ -race` green · `tsc --noEmit` clean · vitest 96 files / 1362 tests passed · eslint unchanged at baseline — this branch touches **zero** files under `web/`.

## Out of scope, flagged not fixed

Two costs the motif work does not own and this PR does not touch, both measured in the same lab:

- `GET /lenses/all/search` — 470 ms (lab), 97 ms on a live server. The embedding/RRF path.
- `GET /lenses/all/stats` — ~86 ms (lab), 83 ms live.

After this fix, `/lenses/all/stats` is the only large term left in a lens-open request set (156 ms → 112 ms wall for the six requests the summary panel fires). Whether to pursue either is a scope call for the maintainer.

`useMotifClusters` still fires one cluster-detail request per motif on an opened fact. At 8.3 ms per detail rather than 105 ms it is no longer the user-visible problem; batching it is a separate judgement call, deliberately deferred.

### DIFF
--- a/internal/textnorm/stem_memo_bench_test.go
+++ b/internal/textnorm/stem_memo_bench_test.go
@@ -1,0 +1,77 @@
+package textnorm
+
+import "testing"
+
+// The measurement behind Stem's memo. Run:
+//
+//	go test ./internal/textnorm/ -run XXX -bench 'Stem|Tokens' -benchtime 2000x
+//
+// On an Apple M4 (2026-09-01), rounded to the precision that reproduces —
+// the warm figures are tens of nanoseconds and move run to run, so read the
+// ORDER, never the digit, and do not treat a differing last digit as a
+// regression:
+//
+//	BenchmarkStemCold      ~25 µs/op    ~480 B/op   13 allocs/op
+//	BenchmarkStem         tens of ns      0 B/op    0 allocs/op
+//	BenchmarkTokensCold    ~45 µs/op   ~1450 B/op   35 allocs/op
+//	BenchmarkTokens      hundreds of ns  128 B/op    2 allocs/op
+//
+// Three orders of magnitude on Stem is the result; the allocation counts are
+// the part worth pinning exactly, and the hit path's zero is the one a change
+// here must not move.
+//
+// Cold is what EVERY call cost before the memo, not a rare first-call price:
+// nothing cached the pluralizer's answer, so the knomit_motif_key SQL callback
+// paid it per token per motif per row on every request. End to end that was a
+// five-mount lens vocabulary at ~133ms; with the memo it is ~9ms.
+//
+// The cold benchmarks clear the memo per iteration OUTSIDE the timer, so they
+// measure go-pluralize and not map churn. One of Cold's allocations is the
+// strings.Clone that keeps an entry from pinning its caller's buffer; it lands
+// on the miss, and the hit path stays at zero allocations.
+
+var benchToken = "vulnerabilities"
+
+var benchCanonical = Canonicalize("failure-presents-as-success")
+
+var stemSink string
+
+var tokensSink []string
+
+func BenchmarkStem(b *testing.B) {
+	stemSink = Stem(benchToken) // warm
+	b.ReportAllocs()
+	b.ResetTimer()
+	for b.Loop() {
+		stemSink = Stem(benchToken)
+	}
+}
+
+func BenchmarkStemCold(b *testing.B) {
+	b.ReportAllocs()
+	for b.Loop() {
+		b.StopTimer()
+		resetStemMemo()
+		b.StartTimer()
+		stemSink = Stem(benchToken)
+	}
+}
+
+func BenchmarkTokens(b *testing.B) {
+	tokensSink = Tokens(benchCanonical) // warm
+	b.ReportAllocs()
+	b.ResetTimer()
+	for b.Loop() {
+		tokensSink = Tokens(benchCanonical)
+	}
+}
+
+func BenchmarkTokensCold(b *testing.B) {
+	b.ReportAllocs()
+	for b.Loop() {
+		b.StopTimer()
+		resetStemMemo()
+		b.StartTimer()
+		tokensSink = Tokens(benchCanonical)
+	}
+}
--- a/internal/textnorm/stem_memo_test.go
+++ b/internal/textnorm/stem_memo_test.go
@@ -1,0 +1,211 @@
+package textnorm
+
+import (
+	"strings"
+	"sync"
+	"testing"
+	"unicode"
+	"unsafe"
+
+	"github.com/gertd/go-pluralize"
+	"github.com/stretchr/testify/require"
+)
+
+// resetStemMemo empties the memo so a test starts from a known state. Tests
+// that read the memo must not race with a neighbour populating it, so they run
+// serially — none of them calls t.Parallel().
+func resetStemMemo() {
+	stemMemo.Range(func(k, _ any) bool {
+		stemMemo.Delete(k)
+		return true
+	})
+}
+
+func stemMemoEntries() map[string]string {
+	out := map[string]string{}
+	stemMemo.Range(func(k, v any) bool {
+		out[k.(string)] = v.(string)
+		return true
+	})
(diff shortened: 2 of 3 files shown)


knomit:
Knowledge captured from this PR into the knomit KB (knomit-kb, branch `agent/mindev.local-8ef0cd32`, commit `36a9a72f`):

- `kb/gotchas/store/motif/stem-cost-in-sql-callback/fe2b9f0f.md` — textnorm.Stem costs ~25µs per token and the knomit_motif_key SQL callback pays it PER ROW; it only bites on branches motif_aliases does not cover, and that second condition is why it hid
- `kb/decisions/textnorm/stem-memoization/6373ce10.md` — Stem is memoized IN PROCESS as a pure function (not a content-addressed cache, not persisted): a remembered pure answer has no invalidation story at all, and a persisted one could outlive the library that wrote it

PR #185, opened 2026-09-02
perf(lens): make opening a lens fast — stats and search stop multiplying by mount count

knomit:
Opening a lens was slow because three independent things each multiplied by the mount count. On the 5-mount `all` lens, search dropped **483 ms → 121 ms** and stats **88 ms → 52 ms**, with **no change to a single byte on the wire**.

## What was actually wrong

Measured first, in a lab (rsync copy of a real `~/.knomit`, own binary, own port, `lsof`-verified port ownership on every run), before any fix was written. Two of the three causes were not what we expected going in.

**1. The same query string was embedded once per mount.** Instrumented, a 464 ms `/lenses/all/search?q=cache` spent 406.8 ms — **88%** — in five `EmbedQuery` calls of 81.1/81.3/81.4/81.4/81.6 ms. The tell is that it is a fixed per-mount cost with no relation to corpus size: the same query against an **empty** mount (0 facts) also cost 82.7 ms, while that same mount answered a text-less filter in 0.42 ms.

`SearchOptions.QueryVec` has always existed for exactly this, and the repo-scoped `/search` handler has always set it. Only the federated twins diverged — they grew a loop the single-mount original did not have. Four call sites had it: web lens search, web lens `facts?q=` (the Library search box, and the worst-hit at 494 ms), and MCP's `queryFirstCall` *and* `queryRecent`.

**2. `/lenses/{lens}/stats` fanned out twice and threw away most of the second pass.** The second fan-out is there for a real reason — `store.Stats` cuts each mount's top-N by that mount's *own* default axis, so when the caller names no axis and the mounts disagree, every mount's highlights were cut by a different ruler than the union ranks by. Pinning them to the resolved pooled axis and re-fetching is correct and stays. What it re-fetched was not: it called the whole of `Stats` again and read only `Highlights`/`HighlightsFallback` out of it.

```
pass 1  per-mount Stats + Activity   43.2 ms
pass 2  the same Stats again         37.0 ms   ← 44% of an 84 ms request
        of which highlights          10.6 ms   ← all it wanted
```

This is not a rare edge case. Mounts disagreeing is the normal state — core and agentic-engineering default to `impact`, knomit-kb, knomit-io-kb and test to `confidence` — and the UI sends no axis. Every lens open paid it.

**3. Both web lens fan-outs were serial loops**, so a lens read cost the *sum* of its mounts. Nothing required that: mounts are independent SQLite databases, and `RepoInstance.Acquire` is an RWMutex **read** lock plus a WaitGroup — never exclusive. `internal/mcp/query.go`'s twins have always been parallel; this reuses that pattern rather than re-inventing it.

**Not** the cause, despite being the leading hypothesis going in: `CommitLogActivity` — the `COUNT(DISTINCT commit_hash)` with wall-clock cutoffs — is 0.35–3.5 ms per mount, **6.9 ms total, 8% of the request**.

## Numbers

Base and final binaries run back-to-back against the same home, same machine state. curl n=9, p50.

| endpoint | before | after | |
|---|---|---|---|
| `GET /lenses/all/search?q=cache` | 483.4 ms | 121.1 ms | 4.0× |
| `GET /lenses/all/facts?q=cache` | 493.5 ms | 147.3 ms | 3.4× |
| `GET /lenses/all/stats` | 87.6 ms | 52.4 ms | 1.7× |
| `GET /lenses/all/stats?axis=impact` | 48.7 ms | 42.3 ms | — |
| `GET /lenses/all/facts` (no text) | 5.5 ms | 5.3 ms | — |

`?axis=impact` is the **control**: it skips the corrective pass entirely, so it must *not* move. The ~3.5 ms it wobbles by is the noise floor the other numbers have to clear.

Per commit, on the endpoint each one targets:

| | base | + embed hoist | + highlights-only | + parallel |
|---|---|---|---|---|
| `search?q=cache` | 465 ms | 134 ms | 148 ms | 121 ms |
| `stats` | 84 ms | 86 ms | 67 ms | 52 ms |

### The slope is the point

`repo=`-narrowed fan-outs, 1..5 mounts:

| mounts | search before | search after | stats before | stats after |
|---|---|---|---|---|
| 1 | 113 ms | 119 ms | 27 ms | 29 ms |
| 2 | 209 ms | 120 ms | 69 ms | 43 ms |
| 3 | 303 ms | 121 ms | 81 ms | 48 ms |
| 4 | 387 ms | 121 ms | 84 ms | 50 ms |
| 5 | 470 ms | 123 ms | 85 ms | 52 ms |

Search went from **+~89 ms per mount to +~1 ms per mount** — flat. At the target scale of lenses up to ~40 mounts, that is **~3.6 s → ~110 ms**. The first two commits made the constant small; this one stops it multiplying.

`BenchmarkFanOutMounts` measures the fan-out's own shape without curl or SQLite in the way (M4, 10 cores, `-benchtime 200x`):

| mounts | 1 | 5 | 20 | 40 |
|---|---|---|---|---|
| serial | 0.51 ms | 2.25 ms | 9.02 ms | 18.07 ms |
| concurrent | 0.46 ms | 0.59 ms | 1.23 ms | 2.41 ms |

At one mount the two agree, as they must. Give it a large `-benchtime`: at 20–50 iterations warm-up noise is the size of that one-mount difference and reads as a speed-up that is not there.

## Semantics

14 lens reads — every axis, repo-narrowed, path-scoped, searches with limits and similarity floors, text and text-less facts browses, a filter-only search — captured from the pre- and post-change binaries against the same home. **14 of 14 byte-for-byte identical**, with no normalisation: the wall-clock activity counters did not tick between captures, so nothing had to be excused.

**Two stated limits of that evidence**, because "14 of 14" reads stronger than it is:

1. **A single capture cannot falsify nondeterminism.** Concurrency's failure mode is an answer that varies between runs, and one capture per binary samples that once. The evidence for determinism is not the captures — it is the forced-delay tests, which invert completion order relative to binding order (slowest mount first; the write mount slowest *and* sharing a path with a read mount that finishes first) and pin the result anyway.
2. **The 5-mount lab home does not exercise the fork-shared-UUID dedupe path.** `WriteFirstWinners` exists for a re-rooted fork mounted beside its upstream — distinct root-commit IDs sharing server-generated fact UUIDs — which this corpus has no instance of. That path is covered by unit tests, not by these captures.

Invariants held throughout: RFC §9.1 (any mount error fails the whole request), `count-vs-transfer` (`total` counts matches, never fetched rows), qualified highlight paths (`lensWirePath` after the dedupe, untouched), no federation metadata added, no new constants, and **no caching of any kind** — every fix is algorithmic.

## Concurrency, specifically

Fetch in parallel, **reduce in order**. Every goroutine writes only its own slot; the reduction runs afterwards on the request goroutine, walking targets in binding order. Nothing appends from inside a goroutine — a mount's *index* is its identity to `FuseRRF` and `WriteFirstWinners`, so collecting results as they arrive would re-map every row's mount, changing both the fused order and which copy of a cross-mount duplicate wins.

§9.1 is preserved and made **deterministic**: when several mounts fail, the reported error is the **lowest-indexed** one — what the serial loop would have hit first. Racing goroutines would otherwise make the message a coin-flip, and a lens error names the offending mount's branch. Pinned with a fast-failing high-index mount racing a slow-failing low-index one, asserted over ten runs.

A panic in a per-mount goroutine becomes that mount's error rather than the process's death: `net/http` recovers panics on the *request* goroutine only.

## Review provenance

A reviewer pass mutation-verified the concurrency tests (append-from-goroutine, first-error-to-arrive, and full-`Stats` reverts all die on the right assertions), confirmed the hoisted `QueryVec` is safe as a shared read-only slice, and verified all four production providers are stateless. Five items came back and are folded in.

One was a **live crash**, not a style note. `fanOutMounts` built its recover message from `t.RT.RI.Name()`, which reads a field and so panics on a nil `RepoInstance` — the same archive/shutdown race the recover exists to survive. The handler failed on precisely the input it was written for, and the second panic escapes the goroutine and kills the process. Verified by reverting the one line under the new test: it prints `panic: mount store vanished [recovered] / panic: runtime error: invalid memory address or nil pointer dereference` and takes the test binary with it.

The others tightened evidence rather than code: the `sort=recent` MCP hoist was unpinned (and its pre-fix arm confirms it genuinely re-embedded per mount); one test's doc comment claimed more than the test showed; one was vacuous if its fixture's two axes ever ordered alike.

## Deliberately not done

- **`/lenses/{lens}/facts` keeps its serial fan-out.** Its loop re-fans at doubling depth against a merge horizon, and folding that in would have put the one intricate case in the commit a reviewer most needs to read plainly. It got the embed hoist (494 → 147 ms) but not the parallelism; 5.3 ms at 5 mounts.
- **`/lenses/{lens}/topics` keeps its serial fan-out — measured, not assumed.** 0.43–0.60 ms at 5 mounts across the root listing, mid-tree nodes, and a leaf-bearing node where the per-leaf `GetByPath` fires; slope ~0.02 ms per mount, ~1 ms extrapolated to 40. Its second store loop is also **not a per-mount fan-out at all**: it is one `GetByPath` per *winning leaf*, so that cost tracks leaf count, not mount count, and parallelising the mount loop would not touch it.
- **The remaining stats slope is unattributed.** 8 concurrent `/lenses/all/stats` take 717 ms against 105 ms for one, and 4 concurrent reads of the *same* repo (89 ms) do worse than 4 across *different* repos (55 ms) — pointing at the per-repo SQLite pool (`SetMaxOpenConns(4)`) plus CPU. The curl harness that produced those carries ~15–20 ms of process-spawn noise and cannot separate the two. Wants a benchmark over real stores.

One correction worth recording: an earlier writeup attributed the parallel-request cost to "per-mount read locks". That is wrong — `Acquire` is an `RLock`, so concurrent readers cannot serialise on it, which is also why parallelising was safe to do at all.

## Verification

- `go test ./...` green.
- `go test -race` green on `internal/web`, `internal/mcp`, `internal/store`.
- `web/` untouched, so its tsc/vitest/eslint baseline is unaffected.
- Rebased onto `dev` after #182/#184 merged; both conflicts were additive handler signatures. Widening runs before the fan-out, the `QueryVec` hoist stays, the parallel fan-out wraps the same §9.1 semantics. Lab numbers re-verified on the rebased tree (search 120.3 ms, stats 52.4 ms, motifs 8.4 ms).

### DIFF
--- a/internal/mcp/helpers.go
+++ b/internal/mcp/helpers.go
@@ -1,7 +1,10 @@
 package mcp
 
 import (
+	"context"
 	"errors"
+
+	"github.com/rs/zerolog/log"
 
 	"knomit/internal/repos"
 	"knomit/internal/store"
@@ -52,4 +55,36 @@ func storeIndices(ri *repos.RepoInstance) (mcpStore, func(), error) {
 		branches:    svc.Branches(),
 		motifs:      svc.Motifs(),
 	}, release, nil
+}
+
+// fanoutQueryVec embeds a federated query's text ONCE, for every mount in the
+// fan-out to share, and is the reason a lens query is not N inferences of the
+// same string.
+//
+// Without it the cost is real and it is the whole request: store.Search embeds
+// whenever SearchOptions.QueryVec is empty (internal/store/search_query.go), so
+// an N-mount fan-out ran the identical ~81 ms ONNX inference N times. It is a
+// fixed per-mount cost, independent of corpus size — the same query against an
+// EMPTY mount also cost 81 ms, while that mount answered a text-less filter in
+// 0.4 ms. A single-repo binding never showed it: there is only one mount to pay
+// for.
+//
+// The hoist cannot change a result. repos.Manager installs ONE embedder on
+// every repo instance, so the N vectors were already identical by construction
+// — this computes one of them instead of N.
+//
+// A nil vector is the DEGRADED path, not an error: with no embedder, no text,
+// or a failed inference, each mount falls back to exactly what it does today
+// (its own embedder, else keyword-only search). Never fail a query because it
+// could not be embedded.
+func fanoutQueryVec(ctx context.Context, emb store.Embedder, text string) []float32 {
+	if text == "" || emb == nil {
+		return nil
+	}
+	vec, err := emb.EmbedQuery(ctx, text)
+	if err != nil {
+		log.Warn().Err(err).Msg("federated query: embed query failed")
+		return nil
+	}
+	return vec
 }
--- a/internal/mcp/query.go
+++ b/internal/mcp/query.go
@@ -171,7 +171,17 @@ type pagedRowState struct {
 
 // QueryHandler returns the handler function for knomit_query.
 // The repo is resolved from the request context at call time via RepoMiddleware.
-func QueryHandler() func(context.Context, mcpgo.CallToolRequest) (*mcpgo.CallToolResult, error) {
+//
+// If an embedder is supplied it is used to embed a text query ONCE for the
+// whole fan-out (see fanoutQueryVec); without one, each mount falls back to
+// embedding the query itself, which is what this handler did before. Variadic
+// for the same reason LearnHandler is: the embedder is optional, and callers
+// that have none (tests, embeddings-disabled builds) pass nothing.
+func QueryHandler(embedders ...store.BatchEmbedder) func(context.Context, mcpgo.CallToolRequest) (*mcpgo.CallToolResult, error) {
+	var emb store.Embedder
+	if len(embedders) > 0 && embedders[0] != nil {
+		emb = embedders[0]
+	}
 	return func(ctx context.Context, req mcpgo.CallToolRequest) (*mcpgo.CallToolResult, error) {
 		ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
 		defer cancel()
@@ -208,9 +218,9 @@ func QueryHandler() func(context.Context, mcpgo.CallToolRequest) (*mcpgo.CallToo
 		}
 
 		if sort == sortRecent {
-			return queryRecent(ctx, b, sWrite, req, pageSize, maxResults, includeBody)
+			return queryRecent(ctx, b, sWrite, emb, req, pageSize, maxResults, includeBody)
 		}
-		return queryFirstCall(ctx, b, sWrite, req, pageSize, maxResults, includeBody)
+		return queryFirstCall(ctx, b, sWrite, emb, req, pageSize, maxResults, includeBody)
 	}
 }
 
@@ -243,7 +253,7 @@ func mountLabel(rt repos.ReadTarget) string {
 // ordered candidate set from RecentFacts (already filtered + committed_at
 // DESC), snapshots it into a session, and serves the first page through the
 // shared resume path so body hydration and pagination match relevance mode.
-func queryRecent(ctx context.Context, b *repos.Binding, sWrite mcpStore, req mcpgo.CallToolRequest, pageSize, maxResults int, includeBody bool) (*mcpgo.CallToolResult, error) {
+func queryRecent(ctx context.Context, b *repos.Binding, sWrite mcpStore, emb store.Embedder, req mcpgo.CallToolRequest, pageSize, maxResults int, includeBody bool) (*mcpgo.CallToolResult, error) {
 	q, err := parseQueryFilters(req)
 	if err != nil {
 		return mcpgo.NewToolResultError(err.Error()), nil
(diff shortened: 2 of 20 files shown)


knomit:
Knowledge captured from this PR into the knomit KB (knomit-kb, branch `agent/mindev.local-8ef0cd32`, commit `dff9c495`):

- `kb/gotchas/lens/search/790a2d19.md` — A federated fan-out embeds the query ONCE; SearchOptions.QueryVec is the seam. Four twins of the repo /search each grew a per-mount embed; the tell is a fixed per-mount cost with no relation to corpus size
- `kb/decisions/lens/stats/c128c02b.md` — The lens stats axis correction is a SECOND fan-out and must fetch only highlights; mounts disagreeing on default_axis is the normal state, not an edge case
- `kb/gotchas/repos/instance/62d74075.md` — RepoInstance.Acquire is a READ lock; concurrent lens reads do not serialise on it (corrects the 2026-09-01 writeup's attribution — the real contention is the per-repo SQLite pool plus CPU, split unmeasured)
- `kb/invariants/web/lens/eab51403.md` — A parallel lens fan-out reports the LOWEST-INDEXED failing mount: binding order, not completion order, and a panic does not outrank a lower-indexed error
- `kb/conventions/web/testing/d24bb8b0.md` — Recording test stubs for lens providers must be mutex-guarded; the provider contract is one call per mount, concurrently — classify stubs by reachability, not blanket-lock

PR #187, opened 2026-09-14
learn: refuse same-subject writes with candidates (read-before-write)

knomit:
Stage 3 of the response to `~/Desktop/knomit-synthesis-prompt-review.md` (Finding 9, Decisions rows 9/9a/9b/9c of `.claude/plans/2026-09-14-distill-prompt-review-fixes.md`).

`knomit_learn` had no read-before-write. Its dedup searches only the incoming fact's **own category directory**, at the `Dedup` floor, over title+body, and never looks at entities — so two sessions recording one event under two different categories never collided, and the corpus grew near-duplicate pairs that only a later review prune would find. This adds a corpus-wide, entity-anchored band *below* the auto-merge floor that refuses the call and hands back the candidates.

## What a caller sees

```
refused: 1 fact in this call shares a subject with a fact already on this branch. Nothing was written.

fact 0: "Enterprise AI spend is plateauing"
  kb/decisions/accepted/ramp/ai-index/4340df4a.md — Ramp AI Index shows adoption flat (shared: Ramp; similarity 0.76)

To update the existing fact call knomit_update on its path. To assert this is a
different fact, resubmit with distinct_from: [<paths>] on this entry.
```

## The band, and why it is not the one the plan first specified

The first draft said `[SimilarTo, Dedup)`. That is **incoherent at every model**, not merely mis-tuned: `tools/calibrate/embeddings.go:100-102` calibrates `Dedup` and `ReflectNovelty` against `docDocSame` (distinct facts in the same category) but `SimilarTo` against `docDocAll`, so the interval had its two endpoints on two different distributions. The nomic `[0.60, 0.92)` looked sane by coincidence. On the shipped model it spans `(0.18, 0.82)` while distinct same-category pairs sit at mean 0.48 / p99 0.77 — nearly every unrelated pair. `SimilarTo` is also, by its own definition, a floor applied *on top of a top-K cap*, never standalone.

The band is now **`(ReflectNovelty, Dedup)`** — `(0.69, 0.82)` on embeddinggemma, `(0.85, 0.92)` on the nomic fallback. Both endpoints are calibrated against the same distribution, and `ReflectNovelty` is already a rejection floor ("too similar to an existing one").

Residual, stated honestly: distinct p99 (0.77) sits above 0.69, so a few percent of genuinely distinct pairs can still land in band. The non-generic-entity anchor and `distinct_from` bound that, and the log events make the rate measurable.

## Fail open, everywhere

Every "no anchor" branch returns nil and refuses nothing: no entities, no sufficiently specific entity, no query vector, no embedder, no calibrated band for the running model, and any infrastructure error. A refusal is a hard block on the caller, so it must never rest on a signal we do not have.

The gate-off exits matter most, and the first version of them was wrong — see B1 in the review section below. They are now keyed on `params.ForModel` rather than on a threshold value that nothing produces, and both are demonstrated load-bearing by deletion.

## Three bugs in the plan's spec, caught before they shipped

1. **`SearchResult.Score` is `cosine * 100`** (`search_query.go:964`) while `params.Thresholds` are 0–1. The spec'd post-filter `Similarity < th.Dedup` would have compared `71.0 < 0.82` and silently refused nothing — a feature that no-ops while every test looks green. Scores are divided by 100 before any comparison.
2. **Entity normalisation is `store.EntityTagMatches`** (`domain_match.go:65`, case-fold only), not in `search_query.go` where the spec pointed. The adjacent `CanonicalizeTag` de-hyphenises and would collapse `AI-Index` and `AI Index`, widening every refusal.
3. **The band is open at the bottom** — the store filters `cosine > MinSimilarity` strictly.

## Commits

1. **`params`: hold every model's calibrated thresholds as data.** `params` knew `DefaultModelID` but not its geometry, which lived inline on the cgo-linked descriptor — so a cgo-free caller had to import ONNX or retype the numbers. `ForModel(id) (Thresholds, bool)` returns them as data; the gemma descriptor now references it exactly as nomic already referenced `Defaults()`. `TestParamsHasNoDependencies` still passes (the rule bans imports, not data).

   The **bool is load-bearing**: a silent `Defaults()` fallback would judge an unknown model id against nomic geometry with nothing red — the same defect as `bridge_motif`'s "UNKNOWN, not defaulted" (M-5), and what `kb/invariants/integrations/hooks/guards-fail-closed` forbids. Presence is a **map-key probe and must stay one**: `th == Defaults()` would report `nomic-v1.5` — a real, registered model whose calibration *is* `Defaults()` — as missing. `TestEveryDescriptorReadsItsThresholdsFromParams` retires the class rather than fixing one model; it is verified falsifiable (an inlined divergent descriptor fails it).

2. **Task 7** — the candidate search. Whole-branch, banded, entity-anchored.
3. **Task 8** — the refusal, `distinct_from`, telemetry.
4. **Task 9** — tool description, server instructions, remember-skill step 1.
5. **Review pass** — reviewer2's ten findings; detailed below.
6. **Re-review fix** — pin the gate-off warning, not just the gate-off outcome.

## Stage ordering

Per `kb/gotchas/mcp/learn/dedup-merge-stage-ordering`, which requires a new stage to state its side of `applyDedupMerge` and cover the merge path: the stage runs **after** the merge (so a folded fact is not also refused) and **before any write**, including before `computeEvidenceWeights` — a refused call should pay for nothing.

The skip set covers **every index the merge touched**, not merely those that "merged": `mergeFacts` inherits title and body from whichever fact won, so for a touched index the precomputed vector no longer describes what would be written.

`dedupEmbed` is hoisted to the handler so both stages share one embed pass; it already ran before any mutation, so the move is behaviour-preserving.

## Generic entities

An anchor must be non-generic: df at or below `synthesize.DFCeiling(n)` = `max(motifDFCeilingFloor, motifDFCeilingPerCent% of n)`, extracted from the inline computation in `bridge_motif.go` so the corpus-size rule has one home. No new constant, and it resolves to a different df on every corpus.

It is reused for its **shape**: the 2%/floor-12 rule was swept on the *motif* axis, and entities are more numerous and longer-tailed. Measured on the core KB (N=2122, ceiling 42): only 15 of 3724 entities exceed it, and they are exactly the big-tech names; the five known collision pairs' anchors sit at df 1–11. The transfer holds at this operating point. Known limit from the same data: the TPU pair names its subject three ways, so its only shared entity is `Google`, which the ceiling excludes — entity anchors miss spelling variance, and prune remains the backstop.

df comes from the existing `TokenDF(…, "entity")` rather than a new method, so there is one definition of df rather than two.

The two folds do **not** fully agree, and the code says so rather than claiming otherwise: `TokenDF` counts over `COLLATE NOCASE` (ASCII-only) while `EntityTagMatches` folds full Unicode, so `Café`/`CAFÉ` is one entity to the match and two to the count. The df lookup is keyed exactly as the count was requested, which confines the disagreement to that non-ASCII case. Reconciling the folds is a follow-up.

## Telemetry — counted, nothing durable

Refusals **and `distinct_from` bypasses** emit a structured zerolog event carrying the refused index, candidate count, anchor entities, the band endpoints in force, and the model id.

Nothing durable, deliberately: learn is a one-shot call with no session and no work-item row, and a refused call must write *nothing* — a counter row or private-state fact would turn an all-or-nothing refusal into a partial commit, the exact failure the contract prevents, and a subscription branch could refuse but not record. Logging the **bypass** matters as much as the refusal: the failure mode of this gate is `distinct_from` becoming a reflex, which is what `kb/decisions/lens/no-write-time-coherence-gate` predicted of an ignorable warning. Explicitly **not** a throttle.

## Testing

Every band case runs at **both** the nomic fallback and the shipped model's geometry, read from `params.ForModel`. The shared harness (`learn_test.go:57`) pins `params.Defaults()` unconditionally, so tests added to it would silently run at a model knomit does not ship — the same blind spot that let the original band look correct. The near-duplicate guard now probes just above `Dedup` rather than at a comfortable 0.95, so it exercises the real boundary for whichever model is in force.

`internal/mcp` stays free of the cgo parent **by convention** — no archtest enforces it, so the thresholds come from `params`, not `embeddings.Lookup`.

A new angle-based mock embedder places facts on a unit circle so a test can name the cosine it wants; the existing length-based one cannot land a pair *inside* a band.

## Notes for review / post-merge

- `SearchResult.Score` is **polymorphic**: `cosine*100` on the vector path, a literal `100` on the text-less path (`search_query.go:751`), which ignores `MinSimilarity` entirely. Not reachable from this stage (`Text` is title+body, both schema-required) but worth a fact — the next person will otherwise re-derive it from a baffling test.
- `kb/decisions/lens/no-write-time-coherence-gate/fd58e51a.md` (confidence 0.95) rejected a hard-reject gate on learn. This band is distinguishable — it is the *high*-similarity, entity-anchored case, and that decision explicitly keeps dedup as "a different job" — but the two will read as contradictory. Post-merge: add a non-scope clause naming this band (row 9c).
- Other `Thresholds().Return(params.Defaults())` stub sites in the repo are untouched; retiring them is a separate follow-up.

---

## Review pass (commit 5) — reviewer2's ten findings

**B1 — the no-embedder guard could not fire.** The condition was `th.ReflectNovelty <= 0 || th.Dedup <= 0`, and nothing produces a zero `Thresholds`: every descriptor is calibrated and `EmbedderThresholds(nil)` returns the nomic set. The safety was still the accident it was meant to replace — an empty vector slice — and the test passed with both guards deleted. A fair hit, and the irony is that this is exactly the bug class the whole PR is about.

The band now comes from `params.ForModel` keyed by the embedder's own model id, which is what makes "we have no geometry for this" expressible at all, and gives commit 1 a production caller rather than an init-time one. Two structural exits, each logged so a silently-disabled gate is visible.

**Both exits demonstrated load-bearing, as asked:**

```
=== (1) delete the `emb == nil` exit -> TestLearnHandler_NoEmbedderNeverRefused
--- FAIL: TestLearnHandler_NoEmbedderNeverRefused (0.01s)
panic: runtime error: invalid memory address or nil pointer dereference
	.../internal/mcp/learn_same_subject.go:282

=== (2) delete the `ForModel` !ok guard -> TestLearnHandler_UnknownModelTurnsTheGateOff
--- FAIL: TestLearnHandler_UnknownModelTurnsTheGateOff (0.03s)
    Messages: turning the gate off MUST be announced; without the warning
              a disabled gate is invisible

=== restored
ok  	knomit/internal/mcp
```

Exit 2's demonstration was **wrong on the first attempt**, caught in re-review. The original test asserted only that the write landed — and deleting the `!ok` check while keeping the lookup leaves `th` as the zero `Thresholds`, which makes `cosine >= th.Dedup` true for every candidate, so the gate-off *outcome* survives by numeric accident and the test stayed green. My own "proof" had mutated something else (swapping the source for `EmbedderThresholds`, which turns the gate *on*), so it demonstrated a different edit than the comment claimed.

Unlike B1 this was never a dead branch — the guard fires. What deleting it destroys is the `log.Warn`, which is the only signal the gate is off at all. The test now asserts that warning and names the model, keeping the outcome assertion alongside: one pins behaviour, the other pins mechanism. The general form is worth carrying forward — **assert the side effect when the side effect is the value**; a guard whose only unique contribution is a log line cannot be pinned by asserting behaviour that is correct without it.

**B2 — pipeline output.** Two exemptions, both logged as bypasses with distinct messages. A candidate the incoming fact already **cites** is declared lineage and skipped for any origin; `origin` in {distilled, discovered} is exempt outright. The refs limb does the real work, because `Origin` is empty on the struct unless the caller passed it (`learn.go` defaults it at serialize time) — so each limb is tested **alone**, refs-with-empty-origin and origin-without-refs, and neither can pass on the other's behalf. Pipeline prompts untouched.

**B3 — the skill couldn't find what the server refuses.** `knomit_query`'s `entities` filter is AND, but the gate refuses on **any one** shared non-generic entity: querying `[Ramp, AI Index]` hides the existing `[Ramp]` fact that is precisely the overlap. Step 1 now queries on text alone and reads hits for shared entities, and the false claim that it is "the same check server-side" is gone. Both skill copies.

**B4 — whole-call refusal was untested.** Every fixture built a one-fact call, where "never half of it" is vacuous. Now a two-fact call, innocent + colliding, asserting nothing is written and only index 1 is named — and verified falsifiable: a per-fact implementation fails it.

**B5 — `distinct_from` case.** `FactExists` lowercases, so a mixed-case path validated, missed the candidate, and was refused again by a message telling the caller to do what they had just done. Lowercased before comparing.

**Non-blocking, all applied:** fail **open** with a warning on corpus-size / df / search errors (they previously aborted the write, contradicting the doctrine three comments above them and diverging from `applyDedupMerge`); skip candidates whose type is `hypothesis`, since `subsumeHypothesis` settles those in the same commit and `knomit_update` cannot change a type; `LiveFactCount` replaces `Stats().Total`, which bought three `json_each` GROUP BYs and the highlights CTE to return one integer; `gofmt`; refusal-text plural.

**The fold disagreement is corrected, not annotated.** The old comment asserted `TokenDF` and `EntityTagMatches` agree on what one entity is. They do not — `TokenDF` counts over `COLLATE NOCASE` (ASCII-only) while `EntityTagMatches` folds full Unicode, so `Café`/`CAFÉ` is one entity to the match and two to the count. The df lookup is now keyed exactly as the count was requested, confining the disagreement; the comment states the real position. Reconciling the folds is the follow-up.

### Hand-written mock method — read before regenerating

Adding `LiveFactCount` to `FactQuery` broke the checked-in `MockSearchIndex`. `mockgen` cannot run in this tree without adding `go.sum` entries, and this PR deliberately does not touch `go.mod` or `go.sum`.

**`MockSearchIndex.LiveFactCount` (and its recorder) in `internal/synthesize/mock_search_index_test.go` is therefore hand-written, in the exact style mockgen emits, and placed in alphabetical position so a regen produces no diff.** The next `go generate ./internal/store/` run is expected to produce the same two methods and leave the file otherwise unchanged — it is a stand-in for the generator, not a deviation from it. If a future regen produces something different, the generator's output wins.

## Test output

```
$ CGO_ENABLED=1 go test -count=1 ./internal/mcp/ ./internal/store/ ./internal/embeddings/... \
      ./internal/synthesize/ ./internal/fact/ ./test/archtest/ ./tools/bridge/skills/
ok  	knomit/internal/mcp	8.907s
ok  	knomit/internal/store	11.173s
ok  	knomit/internal/embeddings	0.397s
ok  	knomit/internal/embeddings/params	0.727s
ok  	knomit/internal/synthesize	161.936s
ok  	knomit/internal/fact	1.280s
ok  	knomit/test/archtest	2.194s
ok  	knomit/tools/bridge/skills	1.559s

$ CGO_ENABLED=1 go test ./...   # EXIT=0, 50 packages ok, 0 failures
```

History was rebased so each commit passes `go test ./internal/fact/ ./internal/mcp/` on its own (the MN6 allow-list now sits with the commit that exports `DFCeiling`).

### The Dedup two-source split (deliberate, documented, do not "harmonise")

`applyDedupMerge` reads `Dedup` from `EmbedderThresholds`; this gate reads it from `ForModel`. They cannot diverge in production — descriptors read their thresholds from `params`, enforced by `TestEveryDescriptorReadsItsThresholdsFromParams` — and where `ForModel` has no entry the gate is **off**, so no band comparison happens and there is nothing to disagree about. Only a test embedder declaring thresholds its model id does not carry can observe the split.

Moving `applyDedupMerge` onto `ForModel` would be **wrong**: dedup must keep working for embedders `params` does not know, so it would need a fallback — reintroducing exactly the silent nomic default this PR removes. Recorded in the doc comment. Follow-up C if it is ever revisited.


knomit:
## Post-merge knowledge: reviewer's fact

One fact written, deliberately narrower than planned. Recall surfaced a near-duplicate, which is why the plan changed.

**`kb/meta/methodology/testing/sabotage-must-delete-not-replace/101f516e.md`** — *A sabotage that REPLACES a guard instead of deleting it tests a different program — so its green AND its red are both true answers to a question nobody asked.*

### Why this and not the broader fact

The plan called for a methodology fact on "absence represented as a plausible value". Recall found `kb/meta/methodology/green-under-sabotage-has-two-causes` already covers both of this PR's guard failures:

- **Cause 1** (nothing exercises the call site) = B1, the unreachable `th.ReflectNovelty <= 0` branch.
- **Cause 2** (a different line quietly provides the guarantee) = R1, where zero `Thresholds` made `cosine >= Dedup` true for every candidate.

Writing a broad new fact would have duplicated it. What this PR found that is genuinely *upstream* of that fact is that the sabotage itself can be invalid — and both misses here were **replacements, not deletions**, so neither result meant anything.

It also fills a half that fact does not cover: it explains how to read a **green** under sabotage, but miss 2 went **red** for the wrong reason. Red is not self-validating either.

The new fact therefore refs and slightly **corrects** the existing one: its consequence says that on a green you must not conclude "my sabotage was wrong" — right once the mutation is valid, wrong before it. Rule out the invalid mutation first; then its two causes are exhaustive.

### The evidence line worth keeping

With R1's guard deleted, the captured log showed the fact was **still written** (`upsert: using donated embedding`, index sync +1). The outcome was correct without the guard, so no outcome assertion could have caught it even in principle — only asserting the warning fires. Hence the pairing: **assert the side effect when the side effect is the value.**

### Correction to the plan's wording

The validation file records the corollary as *"the mutation must be the literal inverse of the guard's text"*. That is my formulation and it is **wrong** — inverting `if !ok` gives `if ok`, a third program again. @worker2 corrected it and the fact uses theirs: **delete the guard and nothing else, then read why it went red.** The operation is deletion, not inversion, and naming it wrong is precisely what would let the next person repeat the mistake.

Code facts, the `no-write-time-coherence-gate` update, the `SearchResult.Score` polymorphism gotcha, and the 9a cosine measurement are worker2's and will be listed separately.

knomit:
## knomit facts written for this PR (post-merge, at 92097878)

All `src://` refs anchor at the merge commit `920978789f251c550d62f292b9878125b90027d3`.

**Written (5):**

| Fact | Covers |
|---|---|
| `kb/decisions/mcp/learn/same-subject-refusal/f77be375.md` | Decision 9 — the refusal and its full contract: stage placement, whole-call refusal, the two bypasses, the pipeline/hypothesis exemptions, fail-open on every missing signal, telemetry-not-durable |
| `kb/decisions/mcp/learn/same-subject-band-endpoints/0c5a0529.md` | Decision 9a — why `(ReflectNovelty, Dedup)`, and why `[SimilarTo, Dedup)` was wrong at *every* model (endpoints on two different distributions) |
| `kb/decisions/mcp/learn/generic-entity-anchor-ceiling/704f442b.md` | Decision 9b — the generic-entity ceiling reuses the motif df derivation for its *shape*, with the measured numbers and the fold disagreement |
| `kb/decisions/embeddings/params-holds-every-model-geometry/321e8453.md` | The `params` threshold move, why `ForModel`'s bool is load-bearing and must stay a map-key probe, and the deliberate `Dedup` two-source split ("do not harmonise") |
| `kb/gotchas/store/search/score-scale-is-polymorphic/f7f6abc0.md` | `SearchResult.Score` is `cosine*100` on the vector path and a literal `100` on the text-less path, where `MinSimilarity` is ignored |

**Updated (1):**

| Fact | Change |
|---|---|
| `kb/decisions/lens/no-write-time-coherence-gate/fd58e51a.md` | Non-scope clause (row 9c) naming this band as the high-similarity, entity-anchored case that decision explicitly left to dedup — and noting that the "ignorable warning" it feared is still the live risk here, in the shape of `distinct_from` becoming a reflex |

All six are cross-linked via `refs` so the provenance walk connects them.

The methodology fact on absence-represented-as-a-plausible-value is owned by the reviewer, not written here.

---

### ⚠️ Post-merge measurement found a gap in the shipped band

Decision 9a's validation (the five known collision pairs under embeddinggemma) has been run against the real model, with pre-merge bodies read from the core KB's own history. Results in `.claude/plans/2026-09-14-same-subject-band-validation.md`.

| Pair | cosine | vs band `(0.69, 0.82)` | caught? |
|---|---|---|---|
| Coursera Project Helix | 0.8752 | above `Dedup` | **no** |
| Ramp AI Index | 0.7033 | in band | yes |
| Forethought nightwatchman | 0.8384 | above `Dedup` | **no** |
| TPUv7 Ironwood | 0.9179 | above `Dedup` | **no** |

`findSameSubjectCandidates` excludes `cosine >= Dedup` because "a hit at or above Dedup belongs to `applyDedupMerge`". That premise holds only *within a category directory* — `applyDedupMerge` searches `Path: categoryDir`. These pairs are all cross-category, which is why they collided in the first place, so **neither gate fires**: the merge never sees them, and the band excludes them.

As shipped, this catches 1 of the 4 measured pairs it was built for. The other three sit 0.02–0.10 above the ceiling and each carries a clean non-generic anchor (`Project Helix`, `Forethought`, `TPUv7 Ironwood`, all df 0).

Candidate fix, not built pending a ruling: make the upper bound conditional on scope — keep `Dedup` as the ceiling for a candidate inside the incoming fact's own category directory, and apply no upper bound outside it. That restores the stated premise rather than weakening the band, and needs no new threshold.

Also: LiteLLM had no locatable merge commit — reported as not-found rather than approximated. And the pre-merge TPU pair *does* share a non-generic entity (`TPUv7 Ironwood`, df 0), so the "its only shared entity is Google" caveat does not hold for those bodies; `generic-entity-anchor-ceiling` will be corrected once the above is ruled on.

Issue #132, opened 2026-08-25
Self-reference is unchecked across passes: a fact can cite itself in refs, and a candidate can be scored against a cluster containing itself


knomit:
Two instances from the live core drain (operator sessions s71–s80, 2026-08-26), same family as #125 (lineage/self-exclusion missing at enumeration) but in different passes, so filed separately.

1. **A fact citing itself in refs**: `7c5691c8` (neighbour of 7db0d980) lists **its own path** in its refs. Nothing at write or read time rejects a self-referential ref, and downstream consumers that walk refs (provenance, evidence weighting) will count the fact as its own support.

2. **A methodology candidate scored 1.00 against a cluster it is itself a member of** (s75 item 453). The score is genuine — a fact matches itself perfectly — and meaningless. Any scoring pass that compares a candidate against a set must exclude the candidate from the set first.

Both are the same missing invariant as #125's fix (exclude self/lineage at enumeration), applied to refs validation and candidate scoring rather than bridge enumeration. A shared helper ("is X in Y's identity/lineage?") would close all three sites consistently.

knomit:
**Designer ruling (2026-08-26): this issue SPLITS across two phases.** Both halves stay on this issue; only the sequencing differs.

## 132a — a fact citing itself in refs → **P2**, ships now

`7c5691c8` listing its own path in its refs. This is a **write-time validation guard**: the ref gate already resolves and canonicalises every ref, so rejecting a ref equal to the fact's own path is a check in a place that already exists. It needs no lineage index and no design ruling.

Consequence worth restating from the body, because it is what makes this more than tidiness: consumers that walk refs for provenance and evidence weighting will count the fact as **its own support**. A self-citation is the degenerate case of the circular-evidence family this whole cluster is about — one fact, zero hops.

## 132b — a candidate scored against a cluster containing itself → **P3**, gated

s75 item 453: a methodology candidate scoring 1.00 against a cluster it is a member of. The score is genuine and meaningless.

This one needs the shared lineage/member index, because "exclude the candidate from the set before scoring" is the same containment question the index exists to answer — and answering it ad hoc here is exactly how the campaign ends up with five sites each solving it differently. It ships with the other consumers in P3's **third** checkpoint, alongside #125 and #135.

## Why the split rather than holding both

Holding 132a for the structural phase would park a one-check fix behind five design gates. Shipping 132b early would mean building a private containment check that the index then has to absorb — the parallel-implementation shape the corpus already has a fact about.

[...]

PR #188, opened 2026-09-14
prompt(distill): positive test, provenance split, no second methodology threshold

knomit:
Acts on three findings from `~/Desktop/knomit-synthesis-prompt-review.md` — a review of two `knomit_review` sessions on the `core` KB (2026-09-14, 15 distill items → 8 syntheses, 8 declines, 0 retractions).

Prompt text only. **No schema change, no code path change, no new LLM call, no discovery pass.** `effort=normal` still spends zero discovery.

## Finding 1 — the distill prompt had no positive test

Roughly two-thirds of the prompt was prohibitions, against one line of positive instruction ("Identify patterns across these facts"). The reviewing agent imported the `motif_alias` prompt's mechanism test by hand because distill had none — which worked, and is exactly the kind of ad-hoc patch the next agent will not make.

Adds the symmetric test: state in one sentence the mechanism ALL members instantiate, or return no synthesis; put that sentence first in the body. That also leaves a checkable artifact for a later pass. Adds prune's accountability line, which measurably stopped rubber-stamping there.

## Finding 2 — "preserve every caveat" vs "synthesis is compression"

These were in direct contradiction, and with 3–6 members the pair is unsatisfiable. Measured consequence: every synthesis longer than any single member, several longer than all members combined, and **zero retractions across 8 syntheses in both sessions**. The retract limb was dead, making distill monotonically additive — it could only grow the corpus, feeding the duplicate load the next session cleaned up.

Conditions now split by what a consumer would act on:

- **Load-bearing** (compound preconditions, scope qualifiers, "WHAT THIS DOES NOT MEAN") — carried verbatim; they exist to block known misreadings.
- **Provenance** (dates, versions, CVEs, affected lists, measurements) — must NOT be copied; cite the member instead.

**On retraction, and why it stays rare.** Near-total non-retraction over a cluster of observations is intended, not a defect being fixed here: distill is additive over observations and duplicate load is prune's job. The prompt says so, because otherwise the model reads its own zero-retraction record as under-performance and starts retracting things it should not.

What the limb needed was a reachable case, not a looser rule. The first draft of this PR narrowed *why* it was dead without making it live — fidelity already forbids the synthesis from copying provenance, so "retract only a member with no unique provenance" exempted every member carrying a date or source, which is nearly all of them. The live case is now named: a member whose provenance is carried by its **refs rather than its body** — typically a prior synthesis or a digest — whose whole claim the new synthesis restates, with those refs carried forward so nothing is orphaned.

The hypothesis rule takes explicit precedence over it. A confirmed hypothesis is neither a restatement nor provenance-free, so the two rules could not otherwise both be obeyed, and an unretracted confirmed hypothesis stays typed as an unconfirmed prediction — which the prompt then bars from every later distill, making it permanently inert. That retraction is a type transition, not subsumption.

## Finding 4 — the methodology ≥ 0.50 literal, at four sites

Retrieval already filters at the per-repo `methodology_min_score` (default 0.15, `internal/config/config.go:231`) and caps at `methodologyTopK = 3` (`internal/synthesize/review_strategy.go:1140`). The prompts then re-thresholded that list at a fixed 0.50 — a corpus-property constant living in a prompt.

Observed scores across 15 distill items: 0.54, 0.47, 0.46, 0.41, 0.37, 0.36, 0.34, 0.33, 0.28, 0.27, 0.26, 0.25, 0.24, 0.22, 0.20, 0.19, 0.16 — **exactly one candidate in 15 items cleared it.** The one that did changed a synthesis from a share-flavoured headline into an existence claim with an explicit prevalence caveat. Sound mechanism, wrong calibration.

**The rule is now rank-based and carries no literal.** Dropping the gate outright would have traded one calibration error for its mirror image: the 0.50 literal **under-read** (1 mandatory read across 15 items), while a universal "read every candidate" **over-reads** (up to `methodologyTopK` reads per item, on distill *and* reflect *and* hypothesize). The rank rule sits between the two — the top-ranked candidate is read unconditionally, the rest are opened on title — so the cost is **one forced read per item** and the measured value, which sat in the top-ranked candidate, is kept.

Triage and judgment are separated explicitly, because the remainder now turns on titles: a title is enough to decide whether to **OPEN** a candidate, never enough to decide whether it **APPLIES**. Both prompts' "titles alone are not enough" would otherwise contradict the rule beside it.

The literal lived at **four sites across three prompts**, not the two the plan first listed:

| site | file |
|---|---|
| 1 | `prompts/large/distill_user.txt` |
| 2 | `prompts/large/reflect_user.txt` |
| 3 | `hypothesize_strategy.go` — workflow step 2 |
| 4 | `hypothesize_strategy.go` — section header |

The pre-existing assertions pinning `"EVERY candidate"` with "prompt must use forcing language, not 'if useful'" are the only surviving record of the earlier decision that letting the model choose whether to read at all is the failure mode. They are **rewritten to pin the new rule and why the universal read was dropped, not deleted to get green.** A new pin on "never enough to decide whether it APPLIES" stops the remainder regressing to title-only judgment.

`reflect_user.txt` was added after checking the code: the reflect methodology path (`review_strategy.go:1218`) uses the same `MethodologyMinScore()` and the same `methodologyTopK` as distill, so it re-thresholds an already-filtered, already-capped list for identical reasons. Both pinned assertions in `methodology_load_test.go` flip to `NotContains`, so reintroducing a prompt-side threshold fails.

## The envelope reserve commit — please judge this one separately

Adding prompt text overran a transport budget, and the first commit on this branch fixes that before any prompt changes land, so **every commit is green in order**.

`pageEnvelopeReserveBytes` reserves the non-facts part of a delivered page (prompt + response schema + paging fields + envelope), and `maxPageFactBytes = maxDeliveredItemBytes - reserve`. Measured:

- envelope on `dev`, before any edit: **8,089 bytes against 8,192 reserved — 103 bytes of headroom**
- envelope after this PR: **9,881 bytes**

The reserve was already one prompt edit from over-promising, and its own comment claimed a worst case of 5,893 bytes — stale by ~2.2 KB. `TestDeliveredPage_EnvelopeFitsItsReserve` exists precisely to catch this ("either prompt file can be edited... without this test the reserve goes stale silently"), and it did.

Raised to 12 KiB, with the comment re-measured and dated.

**Cost, stated plainly: the per-page fact budget drops from 24 KiB to 20 KiB, so distill items page more often. Correctness is unchanged** — `completion_token` paging already covers multi-page items and is the mechanism the review singled out as working. `maxDeliveredItemBytes` is untouched; that is the MCP output cap and a separate concern.

This is not a corpus-property constant under the plan's Global Constraints: it is a measured transport property of the prompt bytes, and the design comment expects re-measurement.

## Housekeeping caught in review

Dropped "at most three entries" from both prompts: it duplicated `methodologyTopK = 3` in prose with nothing tying them, so raising the constant would leave both prompts lying while the suite stayed green.

Repaired three stale "0.50 mandatory-read threshold" references in `internal/store`, including **non-test source** at `methodology.go:85`. All three are prose; no executable change. Worth noting the direction: self-exclusion matters *more* under a rank-based rule, not less — a self-match scores ~1.0 and therefore ranks first by construction, which is now precisely the candidate the judge is told to read unconditionally.

## Not in this PR

- **Finding 3.1** (emit motif carriers before the rest bucket) — already exists as the discover step (`bridge_motif.go`, effort ≥ medium). Why it missed the harness bridge is a read-only diagnosis, tracked separately.
- **Findings 3.2/3.3/6** (rest-bucket honesty, `declined_reason`) — Stage 2, same prompt file.
- **Finding 5** (`candidate_motifs` in the payload) — its own plan.
- **Finding 7** (`bridged` origin) — declined; origin records how a fact was formed at proposal time (`kb/decisions/mcp/origin-by-grouping-method`).
- **Finding 9** (learn read-before-write) — Stage 3, independent files.

## Verification

`go test ./internal/synthesize/ ./internal/mcp/` — both pass, at every commit in order.


knomit:
## Stage 1 review — reviewer1

Verified independently in an isolated worktree at `6a2e9cbb`: `go test ./internal/synthesize/ ./internal/mcp/` passes (178s / 8.7s), and I re-measured the envelope rather than taking the commit message on trust — `distill envelope=9572, reserve=12288, headroom=2716`, reproducing the stated figures exactly.

**Global Constraints all pass.** Targets `dev`; no `Co-Authored-By: Claude` trailer or session URL in any commit or in the PR body; no schema change and no field became required, so `kb/invariants/synthesize/response-envelope` is untouched; single production render site confirmed at `review_strategy.go:1105`.

**The envelope-reserve commit is correct.** `pageEnvelopeReserveBytes` is an MN13 class-2 resource budget — it asserts nothing about any corpus — and its comment names it as a reserve, states the measurement and dates it. `maxPageFactBytes` stays derived (`maxDeliveredItemBytes - pageEnvelopeReserveBytes`), so 24→20 KiB introduces no literal. Ordering it first was right.

Adding `reflect_user.txt` beyond the plan's three sites is justified: `methodologyTopK` is one shared const used by the distill (`:1197`), reflect (`:1240`) and hypothesize (`:469`) paths, all three calling `ri.MethodologyMinScore()`.

---

### 1. [medium-high] The retract limb still cannot fire — the PR's stated goal is not achieved

`distill_user.txt:57` exempts from retraction "a member that carries its own measurement, date, source, or affected list … **because the synthesis deliberately did not copy those**". But line 51 is what guarantees the synthesis did not copy them. The exemption's condition is entailed by the rule above it, so it is self-satisfying: essentially every real fact carries a date or a source, so essentially every member is exempt by construction.

Decision 2 exists because "the retract limb of distill is effectively dead" (0 retractions across 8 syntheses). This rewrites *why* it is dead without making it live. "Zero retractions is the normal and expected outcome" then describes a rule that cannot fire rather than calibrating one that can, and the PR description's "retraction is decoupled from total preservation" overstates what changed.

Worth deciding explicitly: if near-total non-retraction is the intended steady state, say so and drop the decoupling framing. If retraction is meant to fire sometimes, the exemption needs to turn on something other than the presence of provenance.

### 2. [medium] `ONLY` at line 57 contradicts the standing hypothesis-retraction instruction at line 76

Line 76 (`## Hypothesis handling`): "If a hypothesis is confirmed by other evidence in the batch, retract it and include the hypothesis path in the new synthesis fact's refs."

A confirmed hypothesis is neither a strict restatement of the synthesis nor provenance-free, so the later, stricter `ONLY` forbids exactly what line 76 mandates. The prior rule ("only if the synthesis preserves ALL of its conditions") was satisfiable for a confirmed hypothesis; the new one is not.

Consequence chain: confirmed hypotheses stay in the corpus typed as unconfirmed predictions, and line 75 ("Do NOT use hypothesis-type facts as inputs") then bars them from the next distill — they go permanently inert instead of being promoted. Needs an explicit carve-out in the retraction section.

### 3. [medium] Dropping the 0.50 gate universalises the read requirement rather than removing one

Retrieval floors at `MethodologyMinScore` (default **0.15**, `internal/config/config.go:231`) and caps at 3. With the gate gone, the whole retrieved list becomes mandatory reading — on the review's own observed scores (0.54 down to 0.16) the mandatory-read set goes from 1 to ~17 forced `knomit_query` round-trips, and the same change lands on reflect and hypothesize.

A 0.16 candidate is a near-miss of the floor. The new prose "the list is already filtered to what is worth reading" leans hard on a 0.15 default being a meaningful relevance bar. This may well be the right trade, but it is a real cost, opposite in sign to the "only 1 in 15 cleared it" framing, and it is unstated in the PR body. Please state it there.

### 4. [medium] "holds at most three entries" is untethered from `methodologyTopK`

`distill_user.txt:9` and `reflect_user.txt:39` now assert a count to the model. It comes from `methodologyTopK = 3` (`review_strategy.go:1140`) and nothing connects them — no test references both. Raise the const to 5 and both prompts tell the model "at most three" while five render, with the suite green.

Same defect class the PR is removing, one level down. Cheapest fix is deleting the count: "already filtered to what is worth reading" carries the argument without it.

### 5. [medium, forward-looking] Stage 2 will blind the envelope test in the same PR that spends the headroom

`TestDeliveredPage_EnvelopeFitsItsReserve` renders distill via `RenderDistillWorkItem(one, "kb", maxMethodologySection())`. Stage 2's plan says to update Stage 1 tests "to the four-argument form with `false`". Applied blanket here, the test measures the **shorter** branch — the remainder branch is +545 bytes on the plan's verbatim text.

The envelope test needs `remainder=true`, or a third sub-case for it.

### 6. [low] Three stale references to the deleted threshold remain, one of them in non-test source

The four-site inventory missed `internal/store`:
- `internal/store/methodology.go:85` — non-test source comment: "it cleared the 0.50 mandatory-read threshold"
- `internal/store/methodology_test.go:637` — doc comment, same terms
- `internal/store/methodology_test.go:694-695` — `require.Greater(t, selfScore, 0.50, "…must score above the 0.50 mandatory-read threshold…")`

The assertion still passes but now pins a number that means nothing anywhere, with a failure message pointing at a deleted concept. The non-vacuity property it guards is already covered by the `selfSeen` assertion above it.

### 7. [low] Load-bearing conditions are still carried verbatim from *every* member, so length scales with N

The split removes only provenance from the copy set. A 5-member cluster where each member has one scope qualifier and one "WHAT THIS DOES NOT MEAN" clause still mandates 10 verbatim clauses — longer than any single member, which is Finding 2's measured symptom. The new bound ("longer than its members *combined*") is looser than what it replaced. A per-member bound is closer to what a consumer feels.

### Note — Stage 2's envelope cost is ~1.7 KiB, not ~1 KiB

Measured from the plan's verbatim Stage 2 text: prompt delta 1,027 B (remainder branch) + schema delta 650 B = **1,677 B** against 2,716 B headroom, leaving ~1,039 B. It fits, so not a blocker — but the comment's "roughly 1 KiB more" understates it by ~65% and nobody should treat 1 KiB as the remaining budget. Worth correcting in the comment.

---

**Verdict: changes requested.** Findings 1–3 are judgement calls that need a decision rather than a mechanical fix; 4–6 are cheap. Nothing here threatens the envelope-reserve commit, which I'd merge as-is.

knomit:
## Re-review at `8f56c2f0` — reviewer1

All seven findings addressed. `go test ./internal/synthesize/ ./internal/mcp/ ./internal/store/` passes on re-run in an isolated worktree (153s / 8.9s / 10.4s). Global Constraints still pass: base `dev`, no trailer, no session URL.

**Verified rather than accepted:**

- **F1** — the live case is genuinely reachable. "Provenance carried by its refs rather than its body — typically a prior synthesis or a digest" is not caught by the body-carried exemption, so the limb can now fire. Refs carry forward, so nothing is orphaned. `TestDistillPrompt_RetractLimbNamesItsLiveCase` pins it.
- **F2** — carve-out is in the **Retraction** section where the restriction lives, not a backref from the hypothesis bullet, and it states the reason ("type transition, not subsumption") rather than just the exception. That was the thing to get right.
- **F3/R3** — all four sites, `hypothesize_strategy.go:417` and `:480` included. Triage/judgment separated verbatim. The pre-existing assertions were rewritten to pin the new rule with the rationale recorded, not deleted — confirmed by diff.
- **F5** — the direction argument is correct and worth the comment it got: under a rank-based rule a self-match scores ~1.0, therefore ranks first by construction, therefore is the candidate now read *unconditionally*. #132's guard got stronger. Keeping the assertion and rewriting only its justification was the right call over deleting it.
- **Reserve comment** — re-measured independently: **envelope 9,881, headroom 2,407**, both exact. The 1,677 B projection lands at 11,558 leaving ~730 B, also exact. Re-measuring after the R1/R3 rewrites rather than letting my earlier 1,039 figure rot was the right instinct — it had already gone stale.

---

### One item remaining — [low] a vacuous pin

`methodology_load_test.go` pins the forcing word with:

```go
require.Contains(t, content.Prompt, "Always", ...)
```

For **distill** that passes for the wrong reason. `distill_user.txt:79` contains an unrelated *"**Always** use a path under `{{.OntologyRoot}}/meta/reasoning/`"*, so the word is present regardless of the methodology rule.

Proven by sabotage rather than inspection — replacing line 9's `"Always — its title is not enough to judge whether it applies."` with `"Its title is not enough..."`, leaving line 79 untouched:

```
anchor count BEFORE: 2    AFTER: 1
go test -run TestRenderDistillWorkItem_HeaderAppearsWhenSectionPresent
ok  knomit/internal/synthesize    0.494s      <- forcing word gone, pin still green
```

Contrast, same worktree, confirming the other pins do bite:

| mutation | result |
|---|---|
| `"Read the top-ranked candidate above:"` → `"Read the first candidate above:"` | **FAIL** ✅ |
| reflect's only `"Always —"` → `"However,"` | **FAIL** ✅ |
| distill's methodology `"Always"` removed | **ok** ❌ vacuous |

Only the distill site is affected — `reflect_user.txt` has a single occurrence, so its pin is sound.

Fix is one line: pin `"Always — its title is not enough"` instead of the bare word. That phrase occurs once and carries the forcing semantics the assertion is there to protect.

This is the same failure shape the PR has been fixing all the way down — a guard that reads as protection and enforces nothing. Worth closing given the effort spent on the others.

---

### Still open, not ruled on — [low] F7

Load-bearing conditions are carried verbatim from *every* member, so synthesis length still scales with N; the "longer than its members combined" bound is looser than what it replaced. Noting it so it is dropped deliberately rather than by omission. Not a merge blocker.

**Verdict: approve on the one-line pin fix.** Everything else is verified and I have nothing further.

knomit:
## Approved — `8b1c1dfe`

The pin fix is correct, and worker1 was right that my proposed fix carried the trap I had just named one paragraph earlier: `"Always — its title is not enough"` was split across a newline in `reflect_user.txt` (`"…is not\nenough"`, count 0). Pinning it as I suggested would have failed there, and the tempting repair — shortening the pin until it matches the wrapping — is exactly how the vacuous anchor came about. Reflowing the paragraph so the phrase is contiguous, then pinning it, is the right order.

**Verified at final head, mutation-proven rather than accepted:**

```
distill_user.txt: "Always — its title is not enough" x1 ; bare "Always" x2
reflect_user.txt: "Always — its title is not enough" x1 ; bare "Always" x1
```

Removing the forcing phrase from both, leaving line 79's unrelated `"Always use a path under …"` standing:

```
--- FAIL: TestRenderDistillWorkItem_HeaderAppearsWhenSectionPresent
--- FAIL: TestRenderReflectWorkItem_HeaderAppearsWhenSectionPresent
```

The distill mutation left the bare `"Always"` present (x1) and the test failed anyway — that is precisely the condition under which the old pin passed. The anchor now tracks the rule rather than the vocabulary.

`go test ./internal/synthesize/ ./internal/mcp/ ./internal/store/` — all pass (158s / 8.5s / 11.5s). Base `dev`, no trailer, no session URL in commits or PR body.

### On the proposed `assertOccursOnce` helper

Agreed, and agreed it does not belong in this PR. It catches the first defect outright and converts the second from tempting into loud, which is the right pair of properties. The "which pins may legitimately repeat" question is real and deserves its own pass — several existing anchors are common words where a repeat is harmless.

Worth recording when it happens: both defects here are one shape — **an anchor that matches for a reason unrelated to the rule it guards**. A bare word matches an unrelated sentence; a wrapped phrase matches nothing and invites being shortened until it does. Both present as a green suite.

### Open, deliberately

**F7** — load-bearing conditions still carried verbatim from every member, so synthesis length scales with N. Deferred by master, dropped on purpose rather than by omission.

**Approving.** Stage 1 is ready to merge.

knomit:
## Post-merge fact from this PR

Per the project convention (facts for PR'd code land after the merge, listed on the PR):

**`kb/meta/methodology/testing/anchor-binds-spelling-not-rule/4c9711db.md`**

> A test anchor asserted as `require.Contains(rendered, "<word>")` pins a SPELLING, not the rule it guards — a decoy occurrence elsewhere in the same file satisfies it, and the obvious repair (pin the full phrase) fails on line wrapping, where the tempting fix manufactures the original defect.

Written as a **sibling** to `kb/meta/methodology/conformance-whitelist-not-blacklist/ef85a3c0.md` rather than as a new cluster — that fact's claim ("a conformance test that forbids NAMED identifiers pins a spelling, not a property") is the same mechanism on parsed source; this is it on rendered prompt text. Carries the existing motifs `surface-match-misses-meaning` and `guarantee-provided-elsewhere`; nothing minted.

It also records this instance as a **new surface for the second cause** in `green-under-sabotage-has-two-causes` — *"a DIFFERENT line is quietly providing the guarantee you attributed to it."* There the different line was code; here it is prose in the same asset.

Both forms are in one fact, because Form B is the non-obvious half: an anchor matching nothing is an honest red, and the tempting repair — shortening the pin until it matches the wrapping — **manufactures** Form A. That is why "pin the full phrase" is not sufficient advice alone.

The fact names that the reviewer who found Form A then proposed the fix that walked into Form B, and that the implementer caught it. A competent reader taking the tempting path is the evidence that the path is tempting.

Consequence line is the deferred `assertOccursOnce` helper (Follow-up D) — assert exactly once, not at least once — with an explicit carve-out that exactly-once is wrong for anchors that legitimately repeat.

Refs pin `978aa48f` and the three blobs as merged, so the citation resolves to the state where the defect existed and was fixed.

knomit:
Post-merge knomit facts for this PR, written from `dev` at the merge commit `978aa48f` with `src://` refs pinned to that commit and blob.

| fact | what it records |
|---|---|
| `kb/decisions/synthesize/distill/methodology-read-by-rank/ada61851.md` | Decision 4 as shipped |
| `kb/decisions/synthesize/distill/retraction-is-refs-carried/ce8d488b.md` | Decision 2 as shipped |
| `kb/architecture/synthesize/review/page-envelope-reserve/4b665381.md` | the reserve as a measured transport budget |

**What each one is for, and the misreading it blocks:**

- **methodology-read-by-rank** — records that the rule is rank-based with no numeric literal, that it lives at four sites across three prompts, and the observed score distribution (17 values, 0.54 down to 0.16) that showed exactly one candidate in 15 items clearing the old 0.50 bar. It names the misreading explicitly: this is *not* "read every candidate", which was the rejected first draft and the mirror of the bug — the literal under-read, a universal mandate over-reads, the rank rule sits between. It also records why the rewritten `"EVERY candidate"` assertions were rewritten rather than deleted: their message was the only surviving record of the earlier decision that letting the model choose whether to read at all is the failure mode.

- **retraction-is-refs-carried** — records that near-total non-retraction over a cluster of observations is *intended* (distill is additive; duplicate load is prune's job), so nobody later reads the zero rate as a defect and "fixes" it. It also records the trap found in review: the first rewrite narrowed *why* the limb was dead without making it live, because the fidelity rule entails the exemption. And it records the hypothesis carve-out as load-bearing, with the consequence — an unretracted confirmed hypothesis stays typed as an unconfirmed prediction, which the same prompt then bars from every later distill, so the fact goes permanently inert.

- **page-envelope-reserve** — records the consequence a prompt author must act on: *editing a prompt file spends a transport budget*, which is invisible from both sides. It carries the measured drift (comment said 5,893; reality was 8,089, leaving 103 bytes), the cost of raising the reserve (24 KiB → 20 KiB of facts per page, a throughput cost not a behaviour change), and the warning that `maxDeliveredItemBytes` is a separate concern and not the knob to reach for.

Not written here: the vacuous-anchor lesson from this PR's review, which reviewer1 has already recorded as `kb/meta/methodology/testing/anchor-binds-spelling-not-rule/4c9711db.md`.

Stage 2 (Findings 3.2, 3.3, 6) is open as #190 and builds on this branch's prompt.

PR #189, opened 2026-09-14
learn: scope-conditional upper bound — cross-category near-duplicates were caught by nothing

knomit:
Decision **9d** of `.claude/plans/2026-09-14-distill-prompt-review-fixes.md`, following the post-merge 9a measurement on #187. Full results in `.claude/plans/2026-09-14-same-subject-band-validation.md`.

## The defect

`findSameSubjectCandidates` excluded every candidate at or above `Dedup`, justified by *"a hit at or above Dedup belongs to `applyDedupMerge`, which has already folded it."*

That premise holds **only inside the incoming fact's own category directory** — `applyDedupMerge` searches with `Path: categoryDir` and looks nowhere else. So for a cross-category candidate at or above `Dedup`, **neither gate fires**: the merge never saw it, and the band excluded it.

The premise was not wrong about *what* `applyDedupMerge` does. It was wrong about the **scope** over which it does it.

## Measured, which is how this surfaced

Pre-merge bodies read from the core KB's own history (`ReadFact` with `BeforeCommit` at each merge commit) and embedded with the real embeddinggemma model. Band on the shipped model: `(0.69, 0.82)`.

| Pair | cosine | vs band | anchor (df, ceiling 36) | caught before? |
|---|---|---|---|---|
| Coursera Project Helix | 0.8752 | **above `Dedup`** | Project Helix — 0 | **no** |
| Ramp AI Index | 0.7033 | in band | Fable 5 — 12, Opus 5 — 7 | yes |
| Forethought nightwatchman | 0.8384 | **above `Dedup`** | Forethought — 0 | **no** |
| TPUv7 Ironwood | 0.9179 | **above `Dedup`** | TPUv7 Ironwood — 0 | **no** |

The feature caught **one of the four pairs it was built for**. The other three are not near-misses: 0.02–0.10 above the ceiling, each with a clean non-generic anchor.

*(LiteLLM: no merge commit names a LiteLLM pair; four LiteLLM facts are live and mutually distinct. Reported as not-found rather than approximated.)*

## The change

`Dedup` caps the band only for a candidate **inside** the incoming fact's own category directory. Outside it there is **no upper bound**: anything above `ReflectNovelty` carrying a non-generic anchor is a candidate.

`categoryDirOf` is now the single definition of "the scope `applyDedupMerge` searches", called both where the merge *sets* its search scope and where this gate tests whether a candidate falls *inside* it. Two derivations that drifted would put the upper bound somewhere the merge does not actually cover — this defect over again.

## ⚠️ Behaviour change, intended

A cross-category true near-duplicate now produces a **refusal the caller must act on**, where before it was a silent duplicate write. That is the point of the change, not a side effect.

**Effect on refusal rate:** it rises, by exactly the population that previously slipped through — cross-category pairs at or above `Dedup`. These are the *most* confidently-duplicate writes in the corpus, so a refusal here is the least likely to be a false positive; the residual false-refusal risk identified in 9a lives near the *lower* bound, which is unchanged. The existing log events (refusals and both bypass kinds, each carrying band endpoints and model id) are what make the new rate observable rather than assumed.

**Cross-category is refused, never auto-merged.** Merging would have to pick one of the two paths and would silently move a fact out of the category its author filed it under. Refusing hands that choice to the caller: `knomit_update` the existing fact, or assert distinctness with `distinct_from`.

## Tests

One per pair shape, at **both** the nomic fallback and the shipped model's geometry:

- `(b)` same-category candidate at `Dedup` → **not** a candidate (the merge genuinely owns it)
- `(h)` cross-category candidate at `Dedup` → **is** a candidate
- `(i)` cross-category candidate at 0.98 → **is** a candidate
- `TestLearnHandler_RefusesCrossCategoryCollisionAboveDedup` → refused, and **nothing written** (not auto-merged)
- `TestLearnHandler_SameCategoryAboveDedupStillMerges` → merged, not refused
- every existing in-band case unchanged

Plus, from review, the one case where the cap and `touched` **diverge** — and so the only handler-level test the cap can fail on its own:

- `TestLearnHandler_TwoSameCategoryDupsSecondIsWrittenNotRefused` → first merges, second written, **nothing refused**

`TestLearnHandler_SameCategoryAboveDedupStillMerges` turned out **not** to pin the cap: delete the upper bound and it still passes, because `applyDedupMerge` folds that fact, sets `touched[i]`, and the gate skips it before the cap is consulted. The unit test *does* fail (it calls `findSameSubjectCandidates` directly, bypassing `touched`), so this was a test gap rather than a correctness hole — but the cap had no independent handler-level coverage.

The divergence comes from the merge's own `Limit: 1` and `consumed` set: an existing fact absorbs at most **one** incoming fact per call, and a second matcher is *skipped* rather than chained. So with two same-category facts above `Dedup` matching one existing fact, the second reaches the gate **untouched**, and the cap is the only thing between it and a spurious whole-call refusal. Verified falsifiable: with the cap deleted that case refuses at both threshold sets and writes nothing.

Full suite: `EXIT=0`, 50 packages, `gofmt` clean.

## Corrections to the record

- The pre-merge **TPU pair does share a non-generic entity** (`TPUv7 Ironwood`, df 0). The earlier "its only shared entity is Google, so entity anchors miss spelling variance" note generalised from the *post-merge* survivor, which carries three spellings, to the pre-merge bodies, which do not. Spelling variance remains a real limit of the design in general; this pair is not an instance of it.
- The df ceiling is **branch-relative**: 9b measured N=2122 / ceiling 42 on the agent branch, 9a measured N=1810 / ceiling 36 on `main`. Same rule, different denominators. Every df checked sits far from the cutoff either way, so 9b's conclusion is unaffected — but the ceiling is not a fixed number and should not be quoted as one.

`kb/decisions/mcp/learn/generic-entity-anchor-ceiling/704f442b.md` will be corrected once, after this merges, to carry both of those plus the 9d rule together.


knomit:
## knomit facts for this PR (post-merge, at `bfa93f25`)

All three, as ruled — one correction, one new fact, one instance added to an existing fact.

| Fact | Action | Covers |
|---|---|---|
| `kb/decisions/mcp/learn/generic-entity-anchor-ceiling/704f442b.md` | **updated** | Single pass carrying all three changes: the 9d scope-conditional upper-bound rule; the df ceiling is **branch-relative** (N=2122/ceiling 42 on the agent branch vs N=1810/ceiling 36 on `main` — never quote it as fixed); and the TPU correction below. Motif `guarantee-narrower-than-assumed` added alongside `rule-transferred-across-axes`; `sources` → 2. |
| `kb/meta/methodology/testing/shared-premise-defeats-mutation/3f2b7469.md` | **written** | When a suite and the code it tests rest on the same premise, every mutation preserves that premise, so the suite stays green *by construction* — and stays green correctly. Only measurement against real data can disagree. Motif `failure-presents-as-success`. |
| `kb/meta/methodology/green-under-sabotage-has-two-causes/02cfdf3c.md` | **updated** | Second independent instance of **cause 2**, in a different subsystem, with its stated repair rule followed and shown to work. `sources` → 2. |

### The TPU correction

An earlier revision of `704f442b` said the TPUv7 pair's only shared entity is `Google`, which the ceiling excludes, and concluded that pair could never be caught. **Wrong for the pre-merge bodies**, which share `TPUv7 Ironwood` at df 0 — a perfectly good anchor. The error came from reading the *post-merge survivor* (which names its subject three ways) and generalising backwards to the facts that were merged. Spelling variance is still a real limit of an entity anchor in general; this pair simply isn't an instance, and no measured pair currently is.

### Why the cap finding got an instance rather than a fact of its own

`TestLearnHandler_SameCategoryAboveDedupStillMerges` staying green with the cap deleted is **cause 2 of `02cfdf3c` verbatim** — "a different line is quietly providing the guarantee you attributed to it". That fact also already prescribed the repair we applied: *"Reachable but currently redundant → KEEP it, and extend the test to the case where the two diverge."* Which is exactly the `Limit: 1` second-matcher test.

The structural match settled it, not the thematic one: its existing example is a guard redundant except on a zero-item input where both loops run zero times; ours is redundant except on a two-matcher input where `consumed` skips the second. Same divergence-only-at-the-edge shape. Writing a sibling would have split one shape across two facts — so it became a second source instead, and gives that fact a carrier outside `internal/synthesize`, where both its prior examples lived.

Worth recording that the reviewer was **reading from `02cfdf3c`** when they raised the finding. The corpus predicted both the case and its repair before either of us reasoned about it.

### One ref pending

`shared-premise-defeats-mutation` names `kb/meta/methodology/sabotage-must-delete-not-replace/101f516e.md` as its third sibling in the body but cannot **ref** it yet — that fact was written on another agent's branch and doesn't resolve from this one (agent branches don't share writes; consensus comes at merge into `main`). The ref gets added once that branch merges.

PR #190, opened 2026-09-14
distill: honest rest-bucket prompt, recorded declines

knomit:
Acts on Findings 3.2, 3.3 and 6 from `~/Desktop/knomit-synthesis-prompt-review.md`. Follows #188, which shipped the Stage 1 prompt fixes against the same file.

## Finding 3.2/3.3 — the rest bucket asked a question with no honest answer

A `distill-rest` item is the seeds that landed in **no** multi-seed cluster. It is a leftover set, not a group — but the prompt asked it the same "identify patterns across these facts" it asks of a real cluster. On the reviewed session that item held ~60 facts spanning AI funding, security CVEs, a psychology study on punctuation, and interstellar governance, delivered across 3 pages.

There are no patterns across those facts. What the agent actually did — and the system had no record of — was **silently sub-cluster the bucket, synthesize two coherent sub-groups, and ignore the other ~50**. Whenever an agent's best available move is one the prompt does not describe, that is a prompt bug rather than an agent bug.

The remainder branch now says the set is unclustered, asks for coherent **sub-groups**, states plainly that most members belonging to none is expected and that an empty `synthesize` is the common correct answer, and scopes the "share NO motif, skip" rule per sub-group — in a leftover set motifs are shared in scattered pairs, not set-wide. `subgroups_considered` records every sub-group weighed **including the rejected ones**, which is the part that was invisible. Nothing branches on it.

**No new column.** The flag is derived from the work item's `ClusterKey` via `isRemainderItem`: remainder chunks are enqueued as `distill-rest-<chunk>`, and every other distill item is a promoted cluster (`distill-promoted-`) or a RAPTOR follow-up. The key already records this, and a column could disagree with it.

## Finding 6 — declines were indistinguishable from not having looked

`{"synthesize": [], "retract": []}` is identical whether the agent read every member and judged honestly that no mechanism joins them, or skimmed and gave up. Eight of fifteen items declined that way, and at least one was probably wrong — two facts in different topic trees that jointly supported a real claim about containment failing on the infrastructure side rather than the model defeating it. Nobody could see it had been considered.

`declined_reason` (small enum) and `declined_note` (one sentence) are both **optional**: only `synthesize` is probed by `requireResponseKey`, so a response written before they existed still parses — see `kb/invariants/synthesize/response-envelope`. An unknown reason is kept as written rather than rejected, so a future prompt can add a value without a parser change.

**No throttle is built on this.** Prune has a resolution-rate throttle and this deliberately is not one: the point is to make declines visible, and a throttle on an uncalibrated signal would act on noise.

## The completion result now carries a Health line — a deliberate behaviour change

Worth review attention, because **the plan's original wiring rested on two premises that are false of the code**, both verified at `978aa48f`:

1. *"Store the tally on the session in memory."* `store.PipelineSession.Health` is documented "IN MEMORY ONLY — written and read inside the same StartSession call... deliberately not a column", and `pipeline.go` states the contract: "Health descriptors recorded during Plan ride the FIRST result." `recordStats` gives the general rule: "the engine is rebuilt per call, so nothing kept in memory survives to completion." Declines accrue *after* Plan, across later turns — so no turn has a tally both populated and alive.

2. *"Recompute by reading the session's distill items' response column — the store already returns items by session."* The only listing method is `PendingPipelineWorkItems`, which returns items whose `response IS NULL` **by definition**. That is the half with no answers in it. No read path back out of an answer existed at all.

The response column *was* already the durable record; it simply had no reader. So this adds the narrowest one that works — `AnsweredDistillResponses(ctx, sessionID)`, response strings only, distill only, one session — and `completeSession` tallies with `parseDistillResponse` where `synthesize` is empty.

That means the completion result gains a Health line where it carried none. **This is the point, not a side effect:** completion is the only turn on which a decline tally can exist. The line is appended, never assigned, per the health-recorder rule (`TestHealthRecorders_NeverDestroyExistingLines`).

An item that declined without a reason counts as `unstated` rather than being dropped — that item is precisely the case the field exists to surface. A parse failure skips its item instead of failing the session: the line is informational and the session's mutations are already committed, so a summary that errors would be strictly worse than one that undercounts.

## Envelope: measured, not projected

| | bytes |
|---|---|
| reserve (`pageEnvelopeReserveBytes`) | 12,288 |
| distill **remainder** (worst case) | **11,847** |
| distill cluster | 11,301 |
| prune | 6,845 |
| **headroom** | **441** |

`TestDeliveredPage_EnvelopeFitsItsReserve` now runs **a subtest per distill branch**. The remainder ask is the larger prompt, so the blanket `false` the plan first suggested for all call sites would have left the worst case unmeasured while the test stayed green — the stale-reserve failure that test exists to prevent. `TestDistillWorkItem_EveryPageFitsDeliveredCap` measures the remainder branch too, since its page 1 carries the prompt.

441 bytes is thin, and the reserve comment says so: trim prose in `distill_user.txt` before raising the reserve again. The gap against the 730 I projected is a `description` on the `subgroups_considered` schema property, kept deliberately — it helps the model more than the headroom helps us.

## Two fixes from review

**The envelope fixture measured a payload production cannot emit.** The subtest name was being reused as `Item.Type`, which ships on the page, so `"distill-remainder"` inflated the reading by 10 bytes over the production `"distill"`. `stepType` is now a separate field and the numbers above are the production shape.

**A session that declines nothing now says so.** An absent line used to mean any of three things — no distill items ran, some ran and declined none, or the read of the durable record failed — and the failure was the wire-invisible one. The line renders whenever the session had distill items (`distill declines: none of N items`, with no parenthesised breakdown when there is none to show). An absent line now means one thing: no distill work happened.

`TestDistillDeclines_NoLineWhenNothingDeclined` pinned the old rule, so it was **rewritten rather than deleted** — the case it guards is still worth pinning, only the expected output changed, and its message records why the reversal happened.

**A failed read says so too.** `distill declines: unavailable (store read failed)` — still fail-soft and still logged, since the session's mutations are already committed and a summary must not error over a descriptor.

With that, the three causes are fully separated and an absent line has **exactly one meaning: the session ran no distill items.** `TestDistillDeclines_NoLineWhenNoDistillItems` is what pins it.

| session | line |
|---|---|
| declined some | `distill declines: 2 of 2 items (no-shared-mechanism 1, unstated 1)` |
| declined none | `distill declines: none of 1 items` |
| read failed | `distill declines: unavailable (store read failed)` |
| no distill items | *(absent)* |

## Verification

`go test ./internal/synthesize/ ./internal/mcp/ ./internal/store/` — all pass, at every commit in order. The completion-result wiring is mutation-proven: with the `distillDeclineHealth` call removed, `TestDistillDeclines_SurfaceOnTheCompletionResult` fails.


knomit:
## Stage 2 review — reviewer1. Approve, with two low notes.

`go test ./internal/synthesize/ ./internal/mcp/ ./internal/store/` passes on independent re-run in an isolated worktree (150s / 9.2s / 10.7s). Base `dev`, no trailer, no session URL. No migration, no schema column.

### The Task 6 design change — verified, and the plan was indeed unbuildable

I checked both premises at `978aa48f` rather than taking them on report:

1. `res.Health = sess.Health` at `pipeline.go:253` rides the FIRST result, and `recordStats` states the rule outright — *"the engine is rebuilt per call, so nothing kept in memory survives to completion."* An in-memory tally accruing across later turns cannot survive to be rendered.
2. `PendingPipelineWorkItems` is `WHERE session_id = ? AND response IS NULL`. The plan's *"the store already returns items by session"* is true only of the half with no answers in it.

So the plan offered a choice between state that cannot live and a reader that returns the wrong rows. `AnsweredDistillResponses` is the right shape: response strings only, `step_type = 'distill'`, one session — narrow enough that it does not hand back work items the claim protocol has opinions about.

Its load-bearing doc claim also checks out: `Decode`'s distill arm returns `&itemDecision{distill: &result}, response, nil` (`review_strategy.go:847`) — `normalized` **is** the raw agent text, so `parseDistillResponse` applies to the stored strings unchanged.

### Invariant checks — all clean

- `distillResponseSchema`'s top-level `"required"` is still `["synthesize"]`. The three new properties are **not** in it. The `"required": ["members"]` inside `subgroups_considered` is nested in the array item schema and cannot affect a response that omits the array.
- `parseDistillResponse` gains **no** new `requireResponseKey` probe and **no** non-empty check on `synthesize` — correct, since an explicitly empty `synthesize` is exactly the legitimate decline this work instruments.
- `TestParseDistillResponse_DeclinedReasonOptional` exercises the **old** shape `{"synthesize": [], "retract": []}` rather than replacing it, and `DeclinedReason` is a plain `string`, so `TestParseDistillResponse_UnknownDeclinedReasonKept` holds.
- `isRemainderItem` is the only derivation, from `ClusterKey` alone, no column. Negatives use the real `distill-promoted-cluster-3-0` and `raptor-d1-c0-0`.
- MN13: no float literal in `distill_health.go` or the new store method.

### The behaviour change — correctly judged as one

The completion result gaining a `health` line is stated prominently in the PR body as deliberate, with the reasoning. `Health` is `json:"health,omitempty"`, so this is an additive optional key; appended, never assigned. Mutation-proven: removing the `distillDeclineHealth` append from `completeSession` turns `TestDistillDeclines_SurfaceOnTheCompletionResult` **FAIL**, while `TestDistillDeclines_NoLineWhenNothingDeclined` stays green — so the pair discriminates.

Extending `TestDistillWorkItem_EveryPageFitsDeliveredCap` to the remainder branch was your call beyond the plan and it is right: page 1 carries the prompt, so the binding page is the one carrying the largest prompt.

---

### Note 1 — [low] the recorded 11,857 is a fixture artifact; production is 11,847

Measured independently, both branches:

```
distill-cluster    envelope=11301  headroom=987
distill-remainder  envelope=11847  headroom=441
```

The 10-byte gap from the comment's 11,857 is the **subtest name leaking into `Item.Type`**. The table passes `tc.name` as the item type, so the remainder case measures with `Type: "distill-remainder"` (17 chars) where production carries `Type: "distill"` (7). Confirmed by rendering the same prompt twice and varying only that field:

```
Item.Type="distill-remainder"  envelope=11857
Item.Type="distill"            envelope=11847
```

The direction is conservative — the guard is 10 bytes stricter than reality — so this is not a defect in the gate. But the comment presents 11,857 as the measured worst case, and it is the number a future author will trim prose against. Worth either correcting to 11,847 or setting `Type: "distill"` in the table so the fixture matches production.

Same family as `kb/meta/methodology/testing/anchor-binds-spelling-not-rule` from #188: a fixture differing from production in a way that changes what is measured. Benign here; the reason it is worth fixing is that the number is load-bearing for the next person.

### Note 2 — [low] an absent decline line has three causes and one signal

`distillDeclineHealth` returns `""` when there are no distill items, when none declined, **and** when `AnsweredDistillResponses` errors. The read failure logs a `Warn`, but on the wire it is indistinguishable from a healthy session — and wire-invisibility is precisely what this work exists to remove.

Not a blocker, and the fail-soft choice is right. But consider emitting the line whenever `len(responses) > 0` (so "none declined" is stated rather than inferred), or distinguishing the error case. Otherwise the one reader who most needs the signal — someone asking why declines stopped being reported — gets silence.

---

**Approving.** Both notes are low and neither blocks merge.

knomit:
## Approved — `7dd6848d`

Both notes closed. Re-verified independently in an isolated worktree; all three packages pass (162s / 9.8s / 12.2s). Base `dev`, no trailer, no session URL.

**Note 1 — envelope measured in production shape.** Confirmed exact:

```
distill-cluster    envelope=11301  headroom=987
distill-remainder  envelope=11847  headroom=441
```

Splitting `stepType` from the subtest name is the right fix rather than renaming the subtests, and the comment naming the trap — *"a fixture that measures a payload the system cannot emit reports a number about itself"* — is what stops the tempting re-merge of the two fields.

**Note 2 — zero declines stated.** Both render paths verified directly:

```
distill declines: none of 9 items
distill declines: 6 of 9 items (no-shared-mechanism 2, rest-bucket-incoherent 3, unstated 1)
```

No empty parens. `TestDistillDeclines_ZeroIsStatedNotInferred` was rewritten rather than deleted, with the reversal's reasoning in its doc comment — one signal, three causes, the wire-invisible one being the failure. Adding `TestDistillDeclines_NoLineWhenNoDistillItems` to keep the remaining cause pinned was a good call: it preserves the line's meaning as a statement about distill work rather than a stamp on every session.

**Mutation-verified, both directions, with clean discrimination:**

| mutation | result |
|---|---|
| append removed from `completeSession` | `SurfaceOnTheCompletionResult` **FAIL**, `ZeroIsStatedNotInferred` **FAIL**, `NoLineWhenNoDistillItems` pass ✅ |
| zero branch removed | `HealthLine_NoneDeclined` **FAIL**, `ZeroIsStatedNotInferred` **FAIL**, `HealthLine` pass, `SurfaceOnTheCompletionResult` pass ✅ |

Each mutation reddens exactly the tests that assert the mutated behaviour and leaves the others green. That is the property worth having, and it is why these are worth more than their line count.

**Store-read-error path:** recorded as **closed by decision**, not missed. Master scoped this change to the zero-declines case. It remains the one cause of an absent-or-uninformative line that is still wire-invisible, and it is cheap to distinguish if anyone wants it later — but widening an approved PR to get it is the wrong trade.

### On the class, since you raised it

Agreed, and it is the same mechanism. Three instances across #188 and #190, all in tests that read as correct on the page:

- a pin matching an unrelated occurrence elsewhere in the file
- a pin matching nothing because the file wrapped it, where the tempting repair manufactures the first
- a fixture measuring a payload with a field production never emits

The unifying statement in `kb/meta/methodology/testing/anchor-binds-spelling-not-rule` is *an anchor that matches for a reason unrelated to the rule it guards* — and the third widens it past matching to **measuring**. I will add it as a second carrier on that fact rather than a sibling: the mechanism is one, and splitting it would give two facts that each look narrower than the thing they describe. Master has agreed; it lands post-merge with the #190 refs.

**Approving.** Stage 2 ready to merge.

knomit:
## Approved — `f593ac63`

Re-checked only the new commit, as asked. All three packages pass (151s / 8.5s / 11.0s). Commits clean.

**The signal table now has one absent case, and it is pinned.**

| situation | line |
|---|---|
| declined some | `distill declines: 2 of 2 items (…)` |
| declined none | `distill declines: none of 1 items` |
| read failed | `distill declines: unavailable (store read failed)` |
| no distill items | *absent* — pinned by `NoLineWhenNoDistillItems` |

That is what the original note asked for, and the doc comment now states it as a property: *"It returns `""` for EXACTLY ONE reason."*

**Mutation-verified with exact discrimination** — reverting the error branch to a silent `""` fails `TestDistillDeclineHealth_StoreErrorSaysUnavailable` and **nothing else**:

```
--- FAIL: TestDistillDeclineHealth_StoreErrorSaysUnavailable
--- PASS: (the other five)
```

**Fail-soft by construction, confirmed:** `func distillDeclineHealth(...) string` returns no error at all, so no ordering of calls can make a completion fail over a descriptor. That is a stronger guarantee than fail-soft by arrangement, and worth the sentence it got.

### On the stated test limitation — the right call, and I verified the precedent

The test exercises `distillDeclineHealth` directly with a stub rather than end-to-end through `ContinueSession`. I checked the reason rather than accepting it: `pipeline.go:422-428` records exactly this limitation for the answer-write path, marks that branch `UNCOVERED`, and gives the same cause — the engine reaches its indices through `ri.WithRead`, which hands out a concrete `*store.Service` rather than an interface, and `repos.TestInstanceConfig` accepts nothing else, so a fault-injecting `PipelineIndex` cannot be threaded in.

So the limitation is real, pre-existing, and already documented in the same package. Naming it in the test comment rather than building something that *looks* end-to-end is the correct call — a test shaped to look like more coverage than it has is the exact failure this PR pair has now hit three times.

Embedding `store.PipelineIndex` so an unexpected call panics rather than returning a zero value is also right: a loud failure beats a quiet wrong assertion in a stub.

**The missing seam is separate work and bigger than this PR.** Agreed it should not be built here. It is worth recording as a known gap — two branches in one package are now uncovered for the same structural reason, which is the point at which "no seam" stops being a local excuse and becomes a design note.

**Approving.** Stage 2 ready to merge.

knomit:
## Post-merge fact update from this PR

Per the convention (facts for PR'd code land after the merge, listed on the PR):

**`kb/meta/methodology/testing/anchor-binds-spelling-not-rule/4c9711db.md`** — updated, revision `13a444fc`. Not a new fact.

#190's envelope-fixture defect went in as a **second carrier** on the fact written for #188 rather than as a sibling, because it is the same mechanism rather than a neighbouring one. That decision has a cost worth naming: the fact's title was scoped to *matching*, and Form C is about *measuring*, so the title was widened rather than left to under-describe its own contents.

**Form C, as recorded:** `TestDeliveredPage_EnvelopeFitsItsReserve` named its rows `"distill-cluster"` / `"distill-remainder"` and passed `tc.name` as the work item's `Type`. `Item.Type` is delivered on the page and therefore counted, so the remainder row measured a 17-character type where production carries the 7-character `"distill"` — exactly 10 bytes wider than anything the system can emit. Confirmed by rendering the identical prompt twice and varying only that field: **11,857** with the subtest name, **11,847** with the production type.

The direction was conservative, so the gate never under-protected. What was wrong was the number it *reported* — and that number went into the reserve constant's doc comment, which exists to tell the next author how much room they have, with 441 bytes left.

The fact now carries a third `WHAT THIS DOES NOT MEAN` covering exactly that: *"it errs in the safe direction" justifies not treating it as urgent; it does not justify leaving the number.*

**Consequence lines are now two, one per half of the mechanism:**

- binding to text in a rendered artifact → assert the anchor occurs **exactly once**
- **measuring** a delivered artifact → every field that reaches the wire carries its production value, and test-only labels live in fields that do not. *Naming a subtest is not a free act when the name is also passed as data.*

Both refs sets are kept: #188's blobs at `978aa48f`, #190's at `70694835`. Motifs unchanged (`surface-match-misses-meaning`, `guarantee-provided-elsewhere`); nothing minted. `sources` 2 → 3.

**Method note also recorded**, since it is the transferable part: Form C was caught by re-measuring independently and refusing to write off a 10-byte discrepancy from the reported figure as rounding. The gap was the evidence.

knomit:
Post-merge knomit facts for this PR, written from `dev` at the merge commit `70694835` with `src://` refs pinned to that commit and blob.

| fact | what it records |
|---|---|
| `kb/decisions/synthesize/distill/remainder-derived-from-cluster-key/0492e8fa.md` | Findings 3.2/3.3 as shipped |
| `kb/decisions/synthesize/distill/declines-are-stated-not-inferred/0d15c5d2.md` | Finding 6 as shipped, with the four-way signal table |
| `kb/architecture/synthesize/review/completion-result-carries-health/de728d07.md` | the completion result gaining a health line, and the two false premises |

**What each one blocks:**

- **remainder-derived-from-cluster-key** — records that `isRemainderItem` must remain the *only* derivation, and why: the key is already the authoritative record of how the item was enqueued, so a column would duplicate state that can then disagree. It names the limit too, since "encode flags in keys" is the wrong lesson to take away: this works because the enqueue sites are few and named, and a property not already implied by the key needs somewhere real to live. It also carries forward that the remainder ask is the *larger* branch, so it — not the cluster ask — is what a page-budget test must measure.

- **declines-are-stated-not-inferred** — records the four-way signal table and the property that matters: an absent line now means exactly one thing. The instructive part is preserved deliberately — it took three steps to get there (line only when something declined → zero stated → failed read named), and each intermediate state left the signal unable to distinguish its cases. It also notes that "it logs a warning" doesn't count, because the log isn't on the wire the agent reads; the empty-parens rendering trap; and that no throttle is built on this and none should be without calibration.

- **completion-result-carries-health** — the one most likely to save someone a day. Both designs the plan proposed are false of the code, and both sound right: an in-memory tally can't survive the per-call engine rebuild, and `PendingPipelineWorkItems` returns `response IS NULL` **by definition** — the wrong half — which is easy to miss because a by-session listing method visibly exists. It records the working shape (narrow reader over the durable column), the wire-visible consequence to expect, the `Decode`-returns-raw-text dependency a future normalisation would silently break, and the store-fault seam that doesn't exist.

Not written here: the fixture-measurement lesson from this PR's review, which reviewer1 is adding as a second carrier on `kb/meta/methodology/testing/anchor-binds-spelling-not-rule/4c9711db.md`.

Stage 1's facts are listed on #188.

PR #191, opened 2026-09-14
refactor: regroup internal/ by what a package knows (platform tier + fact/textnorm)

knomit:
## What this does

`internal/` had 17 top-level directories, four of which were not subsystems: `memlimit`, `textnorm`, `version` and the already-grouped `obs/`. A reader scanning the tree saw a 5-file cgroup reader sitting next to the 55-file store with no cue that they are different kinds of thing — and nothing in the tree said where the *next* small helper goes, which is how the flat list grew in the first place.

The sorting rule is **group by what a package knows**, which in this codebase is the same axis as dependency weight. The two recorded decisions under `kb/decisions/architecture/internal-package-layout/` already established that nesting must be weight-monotone — a child is never heavier than its parent — and this branch respects both.

| Tier | Knows about | Members |
|---|---|---|
| process | the OS and the binary; nothing about knomit | `crashdump`, `diag`, `logging`, `metrics`, `reqinfo`, `memlimit`, `version` |
| domain leaf | facts as text; no corpus, no git, no SQLite | `textnorm` |
| subsystem | a corpus, a store, a network, a model | everything else — keeps its top-level slot |

Only the first two tiers needed grouping:

- **`internal/platform/`** — the process tier. `obs/` is dissolved rather than nested (`platform/obs/metrics` is three levels for a one-file package, and the five names already describe themselves). Not `internal/runtime/`, which would collide with stdlib `runtime` in every import block needing both.
- **`internal/fact/textnorm`** — its rules were verified against the real knomit corpus and its job is deciding when two fact tags, domains or motifs are the same token. That is a fact concept. textnorm is 75 deps / 9 external against fact's 156 / 29, so the child stays lighter than the parent and the nesting reads honestly.

Deliberately left alone: `refs` (top-level by recorded decision — 388 deps under a 156-dep parent would invert weight), `federate`, `embeddings/params`, `okf/source`, `store/{git,migrate}`, `web/hal`, `synthesize/prompts`, `config`, `app`.

17 top-level entries become 14, and the tree now answers "where does the next one go": knows nothing about knomit → `platform/`; pure text/fact helper → under `fact/`; anything that touches a corpus → its own top-level subsystem.

## The logging inversion

`internal/platform` promises a tier that knows the OS and the binary and nothing about this application. Every package under it held to that already except `logging`, whose `Build`/`BuildWriter` took a `config.LogConfig` — one edge, and therefore the one exception the new archtest would have had to carry. Exception lists are how a layering rule stops meaning anything, so the edge is removed rather than excused: `logging` now declares what it needs via its own `Options` (Format, Level, File and the three lumberjack rotation keys).

`Options` is deliberately **narrower** than `config.LogConfig`. `SlowRequestMS` and `CrashFile` live in that struct too and are read by the HTTP middleware and by crashdump — never by the logger.

The converter itself can live in neither package it names. `platform/logging` must not import `config`; and a method on `config.LogConfig` would give `internal/config` a lumberjack → gzip/flate/crc32 closure that almost every package in the tree would then carry. It lives in **`internal/app`** — the composition tier, which already imports config, already pulls the whole server closure, and is already imported by both binaries that build a logger. `app.LoggingOptions` is the single converter; an earlier revision of this branch spelled the fields out at both call sites, which left a rotation key wired at one and not the other silently dropping in one binary — worse than it sounds, since `tools/desktop` is behind a build tag and so is the copy a default `go build ./...` never looks at.

## The two archtests

`test/archtest/layering_test.go` gains two guards, in the existing style — each doc comment says what failure it prevents, not what it asserts.

**`TestPlatformKnowsNothingAboutKnomit`** walks every package under `internal/platform` and fails on any knomit import outside the tier. The property *is* the tier; those seven packages were grouped because they share it, not because they are small. The failure it prevents is the easy one: reaching for `internal/config` from inside platform compiles and works — `logging` did exactly that until this branch. Stdlib and external libraries are allowed (zerolog, lumberjack and prometheus are the point of several members); no knomit package outside the tier is.

It refuses to pass vacuously twice over: `Fatal` on an empty `go list` result, and `Fatal` if `platform/logging` — the package the guard was written for — is missing from the list. A typo in the pattern would otherwise produce a green test that checks nothing.

**`TestTextnormImportsNoInternal`** pins `internal/fact/textnorm` as a leaf with no knomit dependencies. The edge back to `fact` is already impossible (fact imports textnorm, so it would be an import cycle); what this catches is the *sideways* one to store or config, which would invert the weight ordering that justifies the nesting and charge fact, store and synthesize for it on every normalisation. Deliberately not a zero-dependency check like `TestParamsHasNoDependencies` — go-pluralize and `x/text` are what make it a normaliser rather than a string switch.

Both were sabotage-verified: a real knomit import added to a platform package fails the guard naming both the direct and the transitive edge, and textnorm's guard fails on its own prefix check when given a non-cyclic knomit import.

## Verification

```
gofmt -l <touched files>                    empty
go build ./...                              exit 0
go vet ./...                                exit 0
go test ./...                               exit 0   50 packages ok, 0 FAIL
go test -tags desktop ./tools/desktop/...   exit 0
```

`tools/desktop` is behind `//go:build desktop`, so `go test ./...` skips it — hence the separate run. No production code was changed to make a test pass. Every commit on the branch builds and vets independently.

Beyond the imports, two path references were *live* rather than decorative and would have failed rather than merely misled: `motif_conformance_test.go`'s MN2 target list `require`s `internal/textnorm/textnorm.go` present in a parsed source map, and `stem_memo_bench_test.go` documents the `go test ./internal/textnorm/` invocation for its own benchmark. The Makefile's `VERSION_PKG` ldflags path was re-checked end to end (`knomit version` reports the injected value), not just by eye.

## Follow-ups, not addressed here

Both are **pre-existing gaps** surfaced by review, neither introduced by this branch:

- **`deps()` is silent on parse errors.** `go list -deps` exits 0 on a package containing a syntax error and prints a partial import list; the helper only checks exit status. Not reachable in practice — `go build` and `go vet` fail loudly on the same tree in the same CI job — and it predates this branch, since `TestFactStaysPure` and `TestParamsHasNoDependencies` share the helper.
- **Archtest GOOS and test-import reach.** `go list -deps` resolves for the current GOOS only and excludes `_test.go` imports. `platform/memlimit` and `platform/crashdump` carry per-OS files, so a knomit edge added inside a Linux-only file would pass on a macOS box. The tree was verified clean by hand under darwin, linux and windows, and across all seven packages' test imports — so this is a gap in the guard's reach, not a defect in the tree. Closing it is a few lines setting `cmd.Env` in a loop.

Also unaddressed, cosmetic: three comment lines (`Makefile:33`, `internal/store/domain_match.go:18`, `internal/web/handlers_version.go:11`) now overrun the ~78-col wrap their neighbours keep, because the new paths are longer.


knomit:
Knomit facts written after merge (merge commit b0603592):

**New**
- `kb/decisions/architecture/internal-package-layout/group-by-what-a-package-knows/1a474b97.md` — the sorting rule, the three tiers, the rejected options, the logging.Options inversion, and where the next package goes.
- `kb/gotchas/build/native-libs/worktree-missing-dist/2a1998af.md` — fresh worktrees have no `dist/` native libs, so `go build` fails at link time while `go vet` stays green.

**Updated for the new paths** (old `src://` refs kept, new blob refs at b0603592 added)
- `kb/decisions/build/binary-versioning/4b0a4e47.md` — `internal/platform/version`, `VERSION_PKG`, and the silent-ldflags misreading.
- `kb/decisions/observability/native-only-no-deps/11948a78.md` — `internal/platform/*`, archtest, `internal/obs` gone.
- `kb/decisions/observability/metrics-leaf-registry-and-signals/fec7ffd3.md` — `internal/platform/metrics` (was already stale at `internal/metrics`).
- `kb/decisions/observability/crash-traceback-fd2-bypass/e931ce5c.md` — `internal/platform/crashdump`.
- `kb/decisions/observability/logging-stdout-default-file-optin/73ac51c4.md` — `logging.Options`, `app.LoggingOptions`, the three-place rule for a new `[log]` key.
- `kb/decisions/observability/slow-request-log-detail/faa0b137.md` — `internal/platform/reqinfo`.
- `kb/gotchas/store/motif/stem-cost-in-sql-callback/fe2b9f0f.md` — `internal/fact/textnorm`.
- `kb/gotchas/memlimit/cgroup-v2/namespace-detection/bf9fa487.md` — `internal/platform/memlimit`.

PR #193, opened 2026-09-14
fix(web): ask join-or-subscribe with the Source step's segment control; no push note when subscribing

knomit:
The wizard's Review step asked its one undeducible question — join the branch and push to knomit's own branch, or subscribe and never write — with two 16px browser-default radios crammed under the "What will happen" caption. The Source step asks its own binary (local vs remote) with a segmented control. Two different widgets for two peer choices made the wizard look like two wizards.

This is Option A of three mocks the user picked, because it matches the control the Source step already uses.

## What changed

**One control, two steps.** `segGroup`, `segment(on, tone)`, `segDot`, `segSub`, `segDisclosure` and `type SegTone = 'remote' | 'neutral'` move out of StepSource into `manageStyles.ts`; both steps import them. StepSource's `'local'` tone is renamed `'neutral'` — which is what it always described: the side with no remote to talk about, whether that is "keep it on this machine" or "never write".

**Join leads, subscribe follows.** Join takes the remote tone (#8af, the hue this UI already spends on branches and remote refs — it pushes to the remote's agent branch); subscribe takes neutral slate. Ordering carries the soft preference, never a badge, per `kb/conventions/ui/copy/warning-styling-reserved-for-failures` — a badge on one peer makes the other read as the wrong answer. The Review step now shows two captions, "How to attach" and "What will happen", separated by the same disclosure rule StepSource draws under its own choice.

## Bug this also fixes

`review-write-denied` and `review-write-unknown` rendered whenever the probe had reported on **push** access, regardless of `state.access`. In subscribe mode nothing is ever pushed — the step's own list item 2 says exactly that — so the amber warning landed directly under the line denying it could happen.

Worse, it was not an edge case: the reducer preselects `access: 'subscribe'` when `write_access` comes back `'denied'`, so the contradiction was the **default** rendering of precisely the case the note exists for. Both notes are now guarded on the attach mode. The initialize path (`initialized === 'no'`) always pushes and keeps both notes unchanged.

The refused-push note also becomes a bordered box in CreateRepoWizard's `errText` palette, because it *is* a failure the access check already predicted. The unknown-access note stays plain grey — nothing has failed there, and boxing it would spend the same alarm on the absence of a signal as on a refusal.

## Tests

Three new tests in `StepReview.test.tsx`, two of which were red before this change:

- `says nothing about pushing when subscribing, because nothing is pushed` — pins the bug fix on both notes
- `shows which way is chosen` — `aria-pressed` on `access-join` / `access-subscribe`
- `still warns about a refused push when joining an existing knowledge base` — pins that the guard did not simply delete the warning

One existing test was tightened rather than left silently weakened: `describes a subscription as read-only once it is chosen` scanned the whole pane's `textContent` for "read-only", which the new Subscribe segment's own subtitle now satisfies on its own. It asserts against the list item instead, and was checked for falsifiability — deleting that list item fails exactly that one test.

## Verification

`cd web && npx vitest run && npx eslint . && npx tsc -b` — 1377 tests pass, tsc clean, eslint unchanged from `dev` (94 pre-existing problems on both, none in the four touched files).

Rendered in a browser against a throwaway preview page (since removed) across all four states — join/ok, join/denied, subscribe with a refused probe, and initialize — to confirm the layout, the disclosure rule, and that the subscribe pane shows no push note even with `write_access: 'denied'`.

**Unrelated pre-existing flake, noted so it is not mistaken for this branch:** `Library.keyboard.test.tsx > drives selection/navigation when live` fails roughly 1 run in 8 in the full parallel suite, asserting `NAVIGATE` but receiving `SET_SEARCHING`. Reproduced at the same rate on unmodified `dev` (1 in 10 full-suite runs) and stable when the file runs alone, so it is a load-dependent effect-ordering race in Library's own test, not something this change introduced.


knomit:
Knowledge base updated after merge (knomit-kb, agent branch, commit 96290f6e):

- `kb/decisions/web/create-wizard/review-step-segmented-attach/6d27fca5.md` — why the review step asks join-or-subscribe with the Source step's segmented control (options A/B/C considered, A chosen for consistency with the existing local-vs-remote control; non-scope: not a general radio replacement, radiogroup semantics are known debt to fix in both steps together).
- `kb/incidents/web/create-wizard/push-note-under-subscribe/05a3282d.md` — the push-access notes rendering in subscribe mode, why the reducer's preselect made it the default rendering, the `subscribing` guard, and the two regression tests (including why the second one had to exist).
- `kb/conventions/web/wizard/segmented-choice-control/81405097.md` — the shared `segGroup`/`segment`/`segDot`/`segSub`/`segDisclosure` styles in manageStyles, `SegTone` remote vs neutral encoding state not rank, and the accessibility limitation.

Not filed here: the pre-existing `Library.keyboard.test.tsx` flake noted in the PR body, and the radiogroup-semantics follow-up.

